PR #41, opened 2020-11-13
Resource Loading Errors on Windows

dixonwhitmire:
This PR updates the ResourceReader to utilize InputStreams to resolve issues with loading resources from the classpath on Windows.

Updates include:
- ResourceReader updates to support stream i/o
- Updated GitHub Actions to run tests on Linux and Windows for Java 1.8

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
@@ -9,6 +9,9 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


dixonwhitmire:
@ibmatc @srcarrier - this PR addresses loading resources on Windows

src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
```diff
@@ -31,32 +36,61 @@
...
-        resource = FileUtils.readFileToString(f, StandardCharsets.UTF_8);
+        resource = loadClassPathResource(path.toString());
       }
-      return resource;
-    } catch (IOException e) {
-      throw new IllegalArgumentException("IOexception encountered:" + f.getPath(), e);
+    } catch (IOException ioEx) {
+      String msg = "Error loading resource from " + path.toString();
+      logger.error(msg, ioEx);
+      throw new RuntimeException(msg, ioEx);
```
pbhallam, line 89: any particular reason for throwing RuntimeException instead of IllegalArgumentException?

dixonwhitmire: Nope. I can update this to use an IllegalArgumentException, since that's more descriptive.

Ideally, well at least in my opinion, we would just leave the IOException unhandled and add the "throws IOException" clause to the method. I'm of this opinion since it's still really an IOException that isn't being handled. We are just rethrowing to avoid specifying checked exceptions. I didn't opt for this approach as I didn't want to get into updating how exceptions are handled. This may be something to consider in a future PR.

pbhallam (approved):

PR #43, opened 2020-11-14
Resolve Issue With Loading Default Resources/Configurations From the Classpath

dixonwhitmire:
This PR resolves an issue where the HL7ToFHIRConverter fails to load/instantiate within an application that utilizes the converter as a dependency.

I tested by building and publishing the converter locally, and then referencing it as a dependency within a test application.
I was able to run the code below without errors.

```
        System.out.println(IOUtils.resourceToString("hl7/message/ADT_A01.yml", StandardCharsets.UTF_8, ClassLoader.getSystemClassLoader()));

        HL7ToFHIRConverter converter = new HL7ToFHIRConverter();
        String fhirMessage = converter.convert(hl7Message);
        System.out.println(fhirMessage);
```

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
@@ -7,7 +7,6 @@ package io.github.linuxforhealth.hl7.resource;
 
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
```diff
@@ -54,37 +53,33 @@ private String loadFileResource (File fileResourceConfiguration) throws IOExcept
...
    */
   public String getResource(String resourcePath) {
-    Path path = Paths.get(converterConfig.getResourceFolder(), resourcePath);
+    Path filePath = Paths.get(converterConfig.getResourceFolder(), resourcePath);
     String resource;
 
     try {
-      if (Files.exists(path)) {
-        resource = loadFileResource(path.toFile());
+      if (Files.exists(filePath)) {
```
dixonwhitmire, line 76: If the file exists on the file system we can load using the `filePath`. Otherwise we need to use the resourcePath with "/" to load the resource from the classpath. 

Using the "path" object was incorrect for the classpath use-case since it normalizes a path for the OS which on Windows gives us "\". Additionally the path is to the contents of jar rather than a "typical" file.

pbhallam (approved):

PR #46, opened 2020-11-17
updated loading resources from class loader

pbhallam:


dixonwhitmire (approved):
lgtm we have also tested these changes via a new local jar, so this should be good to merge!

Issue #90, opened 2021-05-10
DiagnosticReport.status: minimum required =1, but only found 0 when validating HL7 ORU-R01 message.


claverols:
**Describe the bug**
Error: "DiagnosticReport.status: minimum required = 1, but only found 0 (from http://hl7.org/fhir/StructureDefinition/DiagnosticReport)" shows when running the current FHIR validator on the converted file from attached HL7 message. 

**To Reproduce**
Run attached HL7 message within latest FHIR validator available and check the output for errors.

**Expected behavior**
The HL7 message
[oru-r01-sample01.txt](https://github.com/LinuxForHealth/hl7v2-fhir-converter/files/6452132/oru-r01-sample01.txt)
 is missing that field, but should the converter be able to deal with it? (ie. map null to Unknown when the expected value is not found?) rather than throw an error?

pydantic.error_wrappers.ValidationError: 1 validation error for DiagnosticReport
__root__ -> status
  field required (type=value_error.missing)

**Desktop (please complete the following information):**
 - OS: MacOS Big Sur
 - Version 11.2.1

**Additional context**

pbhallam:
@claverols I will look into this.

jgrant22:
I don't think I have power to self assign but I will be looking into this card

claverols:
@jgrant22 The FHIR validator tool gets updated quite often; so 1st thing I would download the latest version available and will try to run the FHIR validation against the JSON you will get when converting from the HL7 message using the current HL7toFHIR converter. 

The location for the FhIR validator is: https://confluence.hl7.org/display/FHIR/Using+the+FHIR+Validator#UsingtheFHIRValidator-Downloadingthevalidator

I run the validator with these parameters (-version 4.0 for FHIR version):
java -jar validator_cli.jar -version 4.0 ./../folder/yourconvertedintoFHIR.json -output  ./../folder/outputfromvalidation.json

Search the output from running the FHIR  validation for the reported error. Please let me know if you need more details.

LisaWellman:
DiagnosticReport.status is a required field (https://www.hl7.org/fhir/diagnosticreport.html) so I would expect it to fail validation if it is not set.  The solution to set this field to unknown (https://www.hl7.org/fhir/valueset-diagnostic-report-status.html) if not in the HL7 message seems correct.

cragun47:
In Release 1.0.8

PR #96, opened 2021-06-08
PR for Issue #90

jgrant22:
Error: "DiagnosticReport.status: minimum required = 1, but only found 0 (from http://hl7.org/fhir/StructureDefinition/DiagnosticReport)" shows when running the current FHIR validator on the converted file from attached HL7 message.

Solution: Default to unknown when status is not specified

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
@@ -309,7 +309,7 @@ public class SimpleDataValueResolver {
         if (code != null) {
           return code;
         } else {
-          return null;
+          return "unknown";
         }
 
       };
--- a/src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
+++ b/src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
@@ -122,6 +122,21 @@ public class FHIRConverterTest {
   }
 
   @Test
+  public void test_ORU_r01_without_status() throws IOException {
+    String hl7message = "MSH|^~\\&|NIST Test Lab APP|NIST Lab Facility||NIST EHR Facility|20150926140551||ORU^R01|NIST-LOI_5.0_1.1-NG|T|2.5.1|||AL|AL|||||\r" +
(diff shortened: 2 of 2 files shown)


src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -309,7 +309,7 @@
         if (code != null) {
           return code;
         } else {
-          return null;
+          return "unknown";
```
ssnowski, line 312: Should this be DiagnosticReportStatus.UNKNOWN.toCode() or something similar?

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -121,6 +121,21 @@ public void test_valid_message_but_unsupported_message_throws_exception() throws
...
+            "ORC|NW|ORD448811^NIST EHR|R-511^NIST Lab Filler||||||20120628070100|||5742200012^Radon^Nicholas^^^^^^NPI^L^^^NPI\r" +
+            "OBR|1|ORD448811^NIST EHR|R-511^NIST Lab Filler|1000^Hepatitis A B C Panel^99USL|||20120628070100|||||||||5742200012^Radon^Nicholas^^^^^^NPI^L^^^NPI\r" +
+            "OBX|1|CWE|22314-9^Hepatitis A virus IgM Ab [Presence] in Serum^LN^HAVM^Hepatitis A IgM antibodies (IgM anti-HAV)^L^2.52||260385009^Negative (qualifier value)^SCT^NEG^NEGATIVE^L^201509USEd^^Negative (qualifier value)||Negative|N|||F|||20150925|||||201509261400\r" +
+            "OBX|2|CWE|20575-7^Hepatitis A virus Ab [Presence] in Serum^LN^HAVAB^Hepatitis A antibodies (anti-HAV)^L^2.52||260385009^Negative (qualifier value)^SCT^NEG^NEGATIVE^L^201509USEd^^Negative (qualifier value)||Negative|N|||F|||20150925|||||201509261400\r" +
+            "OBX|3|NM|22316-4^Hepatitis B virus core Ab [Units/volume] in Serum^LN^HBcAbQ^Hepatitis B core antibodies (anti-HBVc) Quant^L^2.52||0.70|[IU]/mL^international unit per milliliter^UCUM^IU/ml^^L^1.9|<0.50 IU/mL|H|||F|||20150925|||||201509261400";
+
+    HL7ToFHIRConverter ftv = new HL7ToFHIRConverter();
+    ftv.convert(hl7message);
+
```
LisaWellman, line 136: It is good that the conversion does not fail, but need to have a check that the status is actually set to "unknown".

Issue #105, opened 2021-06-16
HL7 message without including Patient.gender info (blank PID-8 field) fails to convert


claverols:
**Describe the bug**
An HL7 message without including the patient's gender info (blank PID-8 field) fails to convert. 
This field is optional (Cardinality 0..1 in FHIR 4.0 reference and according to the HL7 reference as well), thus it should convert into a JSON file.

**To Reproduce**
I tried to convert sample_unix.hl7 included in the converter (under src/test/resources) after having removed the gender info from it (|F|), so the PID segment became: 

PID|0010||PID1234^5^M11^A^MR^HOSP~1234568965^^^USA^SS||DOE^JOHN^A^||19800202|||W|111 TEST_STREET_NAME^^TEST_CITY^NY^111-1111^USA||(905)111-1111|||S|ZZ|12^^^124|34-13-312||||TEST_BIRTH_PLACE

I ran the current HL7toFHIR converter on the edited sample_unix.hl7 and I got the error below:
ca.uhn.fhir.parser.DataFormatException: [element=“gender”] Invalid attribute value “?”: Unknown AdministrativeGender code ‘?’
	at ca.uhn.fhir.parser.StrictErrorHandler.invalidValue(StrictErrorHandler.java:50)
	at ca.uhn.fhir.parser.LenientErrorHandler.invalidValue(LenientErrorHandler.java:90)
	at ca.uhn.fhir.parser.ParserState$PrimitiveState.attributeValue(ParserState.java:1350)
...
**Expected behavior**
I would expect the HL7 message to convert without that error, as the gender field is optional.

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Version Big Sur

LisaWellman:
In Release 1.0.9

PR #117, opened 2021-06-24
fix 105 by adding a condition to check for null and unit test

jgrant22:
fix for #105 

```An HL7 message without including the patient's gender info (blank PID-8 field) fails to convert.
This field is optional (Cardinality 0..1 in FHIR 4.0 reference and according to the HL7 reference as well), thus it should convert into a JSON file.```

### CODE REVIEW

File: src/test/java/io/github/linuxforhealth/hl7/segments/Hl7PatientFHIRConversionTest.java [hunk 1]
```diff
@@ -156,4 +157,23 @@ public void patient_multiple_birth_conversion_test() {
...
+            "MSH|^~\\&|MyEMR|DE-000001| |CAIRLO|20160701123030-0700||VXU^V04^VXU_V04|CA0001|P|2.6|||ER|AL|||||Z22^CDCPHINVS|DE-000001\r" +
+                    "PID|0010||PID1234^5^M11^A^MR^HOSP~1234568965^^^USA^SS||DOE^JOHN^A^||19800202|M||W|111 TEST_STREET_NAME^^TEST_CITY^NY^111-1111^USA||(905)111-1111|||S|ZZ|12^^^124|34-13-312||||TEST_BIRTH_PLACE\r";
+
+    Patient patientObjNoGender = PatientUtils.createPatientFromHl7Segment(patientEmptyGenderField);
+    Enumerations.AdministrativeGender gender = patientObjNoGender.getGender();
+    assertThat(gender).isNull();
+
+    Patient patientObjGender = PatientUtils.createPatientFromHl7Segment(patientWithGenderField);
+    Enumerations.AdministrativeGender gen = patientObjGender.getGender();
+    assertThat(gen).isNotNull();
```
LisaWellman, line 176: Let's also add a test for the actual value of the gender (male).

That makes this method a general gender test method, so I would rename patient_gender_test

src/main/resources/hl7/resource/Patient.yml
```diff
@@ -23,10 +23,13 @@ name:
     generateList: true
     expressionType: resource
     specs: PID.5
-gender: 
+gender:
+     condition: $gen NOT_NULL
```
LisaWellman, line 27: I am honestly surprised that we had to do this (that it didn't work before).  We don't want to add checks to every field @pbhallam are we missing something?

jgrant22:
Done

PR #124, opened 2021-07-12
Allergy intolerance criticality

jgrant22:
small fixes for allergy intolerance


src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -260,7 +260,7 @@
         if (code != null) {
             return code;
         } else {
-            return AllergyIntoleranceCriticality.UNABLETOASSESS.toCode();
+            return null;
```
cragun47, line 263: Is there a reason to hardcode null?  I like the option of returning the relative constant.  Just asking... :-)

cragun47 (approved):
Tests need to work.

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
jgrant22: I did null here because if there isn't anything being passed into the field the field should not show. Before if it was empty it said "unable to assess"

src/test/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtilsTest.java
```diff
@@ -86,6 +86,7 @@ public void test_date_diff_valid_values_no_min() {
 
     Long diff = Hl7RelatedGeneralUtils.diffDateMin("2007-11-05",
```
jgrant22, line 87: this test is failing on my other branch as well any suggestions? here's what the test is saying `java.time.DateTimeException: Unable to obtain LocalDate from TemporalAccessor: 2007-11-03T16:33:06.345Z of type java.time.Instant`

src/test/java/io/github/linuxforhealth/hl7/message/Hl7MessageTest.java
```diff
@@ -285,8 +285,8 @@ public void test_allergy_single() throws IOException {
         + "PID|1|465 306 5961|000010016^^^MR~000010017^^^MR~000010018^^^MR|407623|Wood^Patrick^^Sr^MR||19700101|female|||High Street^^Oxford^^Ox1 4DP~George St^^Oxford^^Ox1 5AP|||||||\r"
         + "NK1|1|Wood^John^^^MR|Father||999-9999\r" + "NK1|2|Jones^Georgie^^^MSS|MOTHER||999-9999\r"
         + "PV1|1||Location||||||||||||||||261938_6_201306171546|||||||||||||||||||||||||20130617134644|||||||||\r"
-        + "AL1|1|DA|^PENICILLIN|MO|PRODUCES HIVES~RASH|MO\r" //
-        + "AL1|2|AA|^CAT DANDER|SV";
+        + "AL1|1|DA|^PENICILLIN|MI|PRODUCES HIVES~RASH|MI\r" //
+        + "AL1|2|AA|^CAT DANDER|MI";
```
LisaWellman, line 289: Understand why you modified the "MO" ones, but not sure about this one -- do we have other tests with "SV"?  I would leave this as is.

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -144,6 +144,33 @@ public void test_ORU_r01_without_status() throws IOException {
     assertThat(status).isEqualTo(DiagnosticReport.DiagnosticReportStatus.UNKNOWN);
   }
 
+  @Test
+  public void test_allergy_Criticality(){
```
LisaWellman, line 148: name of test should indicate we are testing that criticality is NOT specified / set, suggest something like test_allergy_no_severity

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -144,6 +144,33 @@ public void test_ORU_r01_without_status() throws IOException {
     assertThat(status).isEqualTo(DiagnosticReport.DiagnosticReportStatus.UNKNOWN);
   }
 
+  @Test
```
LisaWellman, line 147: I would like to see this moved into a Allergy segment test class

LisaWellman (approved):
Would be good to beef up the testcases to include other scenarios, but this covers the issue raised.

src/test/java/io/github/linuxforhealth/hl7/message/Hl7MessageTest.java
LisaWellman: I would like to see this moved into a Allergy segment test class

jgrant22: sv was an error on my end. it is now fixed

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
jgrant22: test case has been moved

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
jgrant22: thanks for the suggestion I made this change

PR #126, opened 2021-07-12
fix mapping for marital status or should we just get rid of mapping

jgrant22:


jgrant22:
there was a problem with martial status, I could map it to something different or is it best to just remove the mapping for C

LisaWellman (approved):

LisaWellman:
Discussed in Patient segment mtg today (Jared, Brian, Anna, Kathryn, I), agreed to change to T

PR #128, opened 2021-07-12
Issue#110 Handle empty telecom correctly

cragun47:
Small change and test to ensure empty telecom information processes correctly -- it should create no telecom section at all.

(There are only four actual lines of code changes in the .yaml.  It looks like more because I added a test, and added hints to the Techniques documentation)

### DIFF
--- a/TECHNIQUES.md
+++ b/TECHNIQUES.md
@@ -4,14 +4,14 @@
 
 Additional information, techniques, and hints about development of templates, java exits, and other enhancements to the converter.
 
-### Getting Java control via data Types
+## Getting Java control via data Types
 
 The converter extracts information via a template and converts to a data type, such as `STRING` and `BOOLEAN`.  You can take advantage of this and create custom data types which can be called for processing of unique inputs.  Before the input is used, your custom data type processor will be invoked.
 
 As an example, type `RELIGIOUS_AFFILIATION_CC` was added to `SimpleDataTypeMapper.java` and mapped to the data handler method `RELIGIOUS_AFFILIATION_FHIR_CC` in `SimpleDataValueResolver.java`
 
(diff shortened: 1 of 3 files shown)


TECHNIQUES.md
```diff
@@ -97,3 +97,36 @@ public static String getAddressDistrict(String patientCountyPid12, String addres
...
+
+```yaml
+telecom_1:
+    condition: $pid14 NOT_NULL    
+    valueOf: datatype/Telecom
+    generateList: true
+    expressionType: resource
+    specs: PID.14
+    vars:
+       pid14: PID.14
```
klwhaley, line 129: Why do we want to do this? instead of the PID.14 NOT_NULL not sure i understand why this is better?

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7TelecomFHIRConversionTest.java
```diff
@@ -61,5 +61,19 @@ public void patient_telcom_test() {
 
   }
 
+  @Test
+
```
klwhaley, line 65: extra line between @Test and the function start

klwhaley (approved):

TECHNIQUES.md
cragun47: Because PID.14 NOT_NULL causes errors in tests.    $pid14 NOT_NULL does not.

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7TelecomFHIRConversionTest.java
cragun47: OK

cragun47:
Updated documentation to be clearer.

cragun47:
@LisaWellman confirms one reviewer sufficient for this one.

PR #129, opened 2021-07-14
use codeableConcept getText function instead of getDisplay

jgrant22:
this will allow us to get the Text directly. originally it was set to get the display. In this case they are the same but I think this is good just in case there is a chance that display and text are not the same.


cragun47 (approved):

src/main/resources/hl7/resource/Patient.yml
```diff
@@ -104,6 +104,7 @@ maritalStatus:
   expressionType: resource
   vars:
     valCodeableConcept: MARITAL_STATUS, PID.16
+    text: String, PID.16.2
```
LisaWellman, line 107: It would be good to get from 16.2 is available, and then 16.9, if available (no need to pull both, just whichever is available, if either are)

jgrant22: fixed

LisaWellman (approved):
Thanks for the 16.9 change.  Would be nice to add a testcase for this scenario, but not requiring it.

PR #136, opened 2021-07-16
fixed the text field for religious affiliation, made changes to marit…

jgrant22:
My thought process was that there's an instance where we can receive just the code. With that, we can get the code, the system and the display. There is also the instance where everything we need will be given in the field. I created marital status to handle both cases but do we want that? I think being able to get as much from just the code is nice. especially if there are a lot of cases where it will just be the code that is given. The only down side is from that code, you can only get the "code, system, display and definition" you cannot get the version or text(unless text is definition?) from that. open to suggestions!!! also added the change to the text field in religious affiliation


src/test/java/io/github/linuxforhealth/hl7/segments/Hl7PatientFHIRConversionTest.java
```diff
@@ -261,20 +261,13 @@ public void patient_gender_test() {
...
     Patient patientObjMarried = PatientUtils.createPatientFromHl7Segment(marriedPatient);
     assertThat(patientObjMarried.hasMaritalStatus()).isTrue();
     assertThat(patientObjMarried.getMaritalStatus().getCodingFirstRep().getDisplay()).isEqualTo(V3MaritalStatus.M.getDisplay());
     assertThat(patientObjMarried.getMaritalStatus().getCodingFirstRep().getSystem()).isEqualTo(V3MaritalStatus.M.getSystem());
 
-    Patient patientObjMarriedAltText = PatientUtils.createPatientFromHl7Segment(AltTextField);
-    assertThat(patientObjMarriedAltText.hasMaritalStatus()).isTrue();
-    assertThat(patientObjMarriedAltText.getMaritalStatus().getText()).isEqualTo("Single");
-    assertThat(patientObjMarriedAltText.getMaritalStatus().getCodingFirstRep().getVersion()).isEqualTo("version 2.5");
-
```
jgrant22, line 279: marriedPatient is setup to support `coding_3` and the AltTextField was set up to test  support for `coding_1` (which is no longer needed)

jgrant22: the only thing is this means we cant get the version anymore as well. Which makes me wonder if we should expect this field to ever give us more than the code?

klwhaley: Lets go with this for now as this is how coding_3 is working.  We will re-evaluate marital status text and version in the future.  I would like you to add back the test for the patient with just S, and the version...so we have it there to enhance in the future.  So can you add it back, and just check we get S, the correct display, system, also check text and version are null, iwth a comment that is as designed for now.

klwhaley (approved):
approved, thanks for handling all the review comments

PR #137, opened 2021-07-16
Issue#131-email-support

cragun47:
Support for email addresses.  Uses value in 13.4 as email address.
This will solve problem of zombie phones seen previously in output.  They were caused by incomplete processing of emails.


LisaWellman (approved):
Could consider a test with a work email.

PR #138, opened 2021-07-16
Do not set CodeableConcept.display from hl7 CWE.2 text

LisaWellman:
Enhanced Allergy tests for this scenario, plus a few more checks.  Note "DRUG" is not valid for AL1-2.

The changes in the OBX test seemed reasonable.  OBX-5 is not a defined format, but in these messages seems to follow the CWE format.  The changes in DifferentObservationValueTest.java are hard to see even with whitespace changes hidden - most of it is the formatter, the real changes are removal of this check in 2 places:
`assertThat(cc.getCoding().get(0).getDisplay()).isEqualTo("No significant change was found");`


cragun47 (approved):
Perfect!

PR #140, opened 2021-07-16
Issue#127 Handle edge conditions for IDs

cragun47:
Improvements to handle edge conditions for ID's:

If the authority has blanks in the name, replace blanks with _'s so the System ID is valid
If the authority is not present in the HL7 message ID's, do not create a System ID in the ID resource
Improve SYSTEM_ID data handler to return null on input of null or empty to support no System ID creation.
Do not assume an authority for an ID created from a PID.19 SSN.

### DIFF
--- a/src/main/resources/hl7/datatype/CodeableConcept.yml
+++ b/src/main/resources/hl7/datatype/CodeableConcept.yml
@@ -43,7 +43,10 @@ coding_4:
        display: CWE.5
        version: CWE.8
 
-text:  
+text:
+     # If hideText is not passed by parent, the value is NULL, making it backward compatible with all other uses of CodeableConcept
+     # Any value in hideText will cause the text to be hidden.  Used by Identifier_SystemID.
+     condition: $hideText NULL 
      type: STRING
      valueOf: CWE.2 |CWE.9 |CE.2 |CNE.2| ST |TX |ID |IS | CWE.5
      expressionType: HL7Spec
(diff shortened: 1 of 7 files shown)


cragun47:
Added to this PR:
    1.  Additional requested tests for email
    2. Cause calculated text from coding to not be shown for IDs.   ("SS", "MR", "NH", "DL", etc.)

src/main/resources/hl7/datatype/Identifier_SystemID.yml
```diff
@@ -9,11 +9,17 @@ type:
    valueOf: datatype/CodeableConcept
    expressionType: resource
    specs: CX.5 | $type
+   constants: 
+      # Because calculated system id's don't have text, force the calculated text from code to be hidden
+      hideText: 'HIDE'
```
klwhaley, line 14: i need to think a bit more about this design to 'hide' the text...but will approve.  We may want to do more on this / change how we have done this...

src/main/resources/hl7/datatype/CodeableConcept.yml
```diff
@@ -43,7 +43,10 @@ coding_4:
        display: CWE.5
        version: CWE.8
 
-text:  
+text:
+     # If hideText is not passed by parent, the value is NULL, making it backward compatible with all other uses of CodeableConcept
```
klwhaley, line 47: i need to think a bit more about this design to 'hide' the text...but will approve. We may want to do more on this / change how we have done this...

cragun47: Acknowledged.

PR #142, opened 2021-07-19
Issue#141-Correct-handle-conflict-multbirths

cragun47:
Changes to correct handling of multiple birth information when only integer or N+integer are present.
See bug description for implemented behavior.

BTW, I did simplify the input and the order of the tests to make them easier to follow.


klwhaley (approved):
Approving, if we figure out complex conditions we can refine

PR #147, opened 2021-07-20
Use \n instead of ~ for multiline reports

LisaWellman:


klwhaley (approved):
Changes look good and reviewed in webex

PR #152, opened 2021-07-26
Upgrade to Junit5

klwhaley:
Update to Junit 5

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ compile group: 'commons-beanutils', name: 'commons-beanutils', version: '1.9.4'
 
 testCompile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
 testImplementation 'org.assertj:assertj-core:3.9.0'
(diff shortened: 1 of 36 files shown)


build.gradle
```diff
@@ -59,8 +59,7 @@ compile group: 'commons-beanutils', name: 'commons-beanutils', version: '1.9.4'
 
 testCompile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
 testImplementation 'org.assertj:assertj-core:3.9.0'
-testImplementation 'junit:junit:4.12'
-  
+testImplementation "org.junit.jupiter:junit-jupiter-api:5.7.0"
```
LisaWellman, line 62: Should we be going to 5.7.2?  That looks like the latest stable release

LisaWellman (approved):

build.gradle
klwhaley: good point, updated! thx

PR #156, opened 2021-07-29
Extensions generic support final

pbhallam:


cragun47 (approved):
The extension improvements look good.  It's too bad we couldn't get the Race handling in, but we'll do that in another.
It is a trivial thing to even mention, but vertical spacing is not consistent.  Certainly no reason to require a change or hold up approval.

xpander53 (approved):

PR #157, opened 2021-07-29
Bjc   extended bi condition tests

cragun47:
Fixed a couple of problems in the BiConditional tests.  Adds full set of tests for biconditions.

### DIFF
--- a/src/main/java/io/github/linuxforhealth/core/expression/condition/SimpleBiCondition.java
+++ b/src/main/java/io/github/linuxforhealth/core/expression/condition/SimpleBiCondition.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2020
+ * (C) Copyright IBM Corp. 2020, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/linuxforhealth/core/expression/condition/SimpleBiCondition.java
```diff
@@ -49,6 +49,11 @@ public boolean test(Map<String, EvaluationResult> contextVariables) {
       ConditionPredicateEnum condEnum = ConditionPredicateEnum
           .getConditionPredicate(this.conditionOperator, variable1.getIdentifier());
       if (condEnum != null) {
+        // if var2 is a string and must be converted to an integer to test
```
klwhaley, line 52: why does only var2 need to be converted? obviously new to this section of code...but should var1 also be converted?

the comment confused me slightly -- we are doing this Only when the condition includes integers... so can we update the comment to clarify that..

src/test/java/io/github/linuxforhealth/core/expression/condition/SimpleBiConditionTest.java
```diff
@@ -11,64 +11,562 @@
...
+    String condition = "$var1 NOT_EQUALS abc";
+    SimpleBiCondition simplecondition = (SimpleBiCondition) ConditionUtil.createCondition(condition);
+    Map<String, EvaluationResult> contextVariables = new HashMap<>();
+
+    assertThat(simplecondition.test(contextVariables)).isFalse();
+  }
+
+  // CONTAINS test one true and one false
+
+  @Test
```
klwhaley, line 118: In the Ends with , contains, starts with etc tests you are testing a variable and a value, never two variables...should we a mix a few of those in?

klwhaley (approved):
Approved. A couple minor comments.

src/main/java/io/github/linuxforhealth/core/expression/condition/SimpleBiCondition.java
cragun47: Will do.

src/test/java/io/github/linuxforhealth/core/expression/condition/SimpleBiConditionTest.java
cragun47: Yes.  I will add those to a new pull.

PR #158, opened 2021-07-30
Iss#149 add review tests, refactor, fix missing bi

cragun47:
This addresses comments in review.
Rather than cut/paste 8 new similar tests, the code was refactored with common routines to reduce code, increase coverage, make future updates are easier, and make what is being tested clearer from looking at code rather than reading the comment.  In the process, discovered two missing comparators:  NOT_EQUALS and NOT_EQUALS_INTEGER.  These were also added.
NOTE:  SimpleBiConditionTest has major structural changes and moving about.   A file compare may be less useful than looking at the complete file.


LisaWellman (approved):
Nice refactor!

Issue #115, opened 2021-06-24
Capture Patient Race as an extension


cragun47:
**Describe the bug**
Create FHIR Patient.extension.patient-race (well-known extension) from PID-10.1

**To Reproduce**
Create a sample HL7V2.6 patient PID segment with aforementioned patient information.

**Expected behavior**
Patient information should be detected and added to corresponding fields in HL7 FHIR resources.

Example FHIR:
```
extension": [{
        "url": "http://ibm.com/fhir/cdm/StructureDefinition/local-race-cd",
        "valueCodeableConcept": {
          "coding": [ {
            "system": "http://terminology.hl7.org/CodeSystem/v3-Race",
            "code": "2028-9",
            "display": "Asian"
          } ],
          "text": "Asian"
        }
      }, {
        "url": "http://ibm.com/fhir/cdm/StructureDefinition/local-race-cd",
        "valueCodeableConcept": {
          "coding": [ {
            "system": "http://terminology.hl7.org/CodeSystem/v3-Race",
            "code": "2106-3",
            "display": "White"
          } ],
          "text": "White"
        }
      }
]
```

**Additional context**
[HL7 FHIR Site](http://hl7.org/fhir/us/core/STU3.1.1/ValueSet-omb-race-category.html)
[Sample results](https://trifolia-fhir.lantanagroup.com/igs/lantana_hapi_r4/MCC-IG/Patient-cc-pat-pnoelle.json.html)

cragun47:
The expected coding has been changed.  The request has been updated.

cragun47:
If the HL7V2 is degenerate and has no coding but only text (happens quite a bit), we interpret that as a code without a coding thus:

```json
{
        "url": "http://ibm.com/fhir/cdm/StructureDefinition/local-race-cd",
        "valueCodeableConcept": {
          "coding": [ {
            "code": "W"
          } ]
        }
      } 
```

LisaWellman:
In Release 1.0.9

PR #162, opened 2021-08-05
Issue #115 simple race built on new extensions

cragun47:
This adds race as an extension as described in #115 

Additional changes to extension to handle repeated `specs` was added by @pbhallam .   Thanks!


LisaWellman (approved):
LGTM

PR #165, opened 2021-08-11
453 evn segment

xpander53:


.gitignore
```diff
@@ -46,4 +46,5 @@ bin/
 .idea
 
 # Local History files
-.history
\ No newline at end of file
+.history
+AnnaConverterFolderTest.java
```
LisaWellman, line 50: why do we need this?  I hope we never delivered this folder!

src/main/resources/hl7/resource/Encounter.yml
```diff
@@ -77,7 +77,10 @@ reasonCode:
     expressionType: resource
     specs: EVN.4 | PV2.3
 
-
+reasonReference:
+    valueOf: datatype/Reference
+    expressionType: resource
+    specs: $Condition | $Procedure | $Observation | $ImmunizationRecommendation
```
LisaWellman, line 83: We do not support ImmunizationRecommenation; I do not think it is anywhere in the converter.  I am surprised you did not get an error, but in any case I would remove.

Similar for Procedure -- though we have that one Jared just added minimal support to get it to work for Identifiers.

.gitignore
xpander53: It's so we can have the test program file in our workspace inline.  But I can remove it np.

src/main/resources/hl7/resource/Encounter.yml
xpander53: I will remove them.

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -353,4 +353,71 @@ private void verifyResult(String json, BundleType expectedBundleType, boolean me
     }
   }
 
+  @Test
+  public void validate_evn_segment () {
```
LisaWellman, line 357: We've been trying to avoid putting tests into this file as it is so general and we don't want it a hug dumping ground.  Maybe create a new EVNSegmentTest in src/test/java/io.github.linuxforhealth.hl7.segments

LisaWellman (approved):
Just a few comments to address, but looks good!

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
xpander53: I created new file  HL7EventTypeFHIRConversionTest and placed them there.

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -353,4 +353,71 @@ private void verifyResult(String json, BundleType expectedBundleType, boolean me
...
+      .collect(Collectors.toList());
+  assertThat(encounterResource).hasSize(1);
+
+  //ENV.4 is used for reasonCode
+  String reasonCode = encounterResource.get(0).getNamedProperty("reasonCode").getValues().get(0).getNamedProperty("text").getValues().get(0).toString();
+  assertThat(reasonCode).isEqualTo("7525");
+
+  //EVN.6 is used for start period if there is no PV1.44 but since we have a PV1.44 it should use that not EVN.6
+  String startPeriod = encounterResource.get(0).getNamedProperty("period").getValues().get(0).getNamedProperty("start").getValues().get(0).toString();
+  assertThat(startPeriod).isNotEqualTo("DateTimeType[2021-03-19T13:47:35+08:00]");
```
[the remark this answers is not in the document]
xpander53: Good idea - I added that.

PR #168, opened 2021-08-19
487 fixes

jgrant22:
this fix covers the bug causing rde_o11 to default to msh.7 (skipping over pv1) and 

adding the system to the patient visit identifier. 
If the segment was ```|7588^^^ACME|```  value would be "7588" and system would be "urn:id:ACME"


src/test/java/io/github/linuxforhealth/hl7/segments/Hl7IdentifierFHIRConversionTest.java
```diff
@@ -226,7 +226,7 @@ public void condition_identifier_test() {
     assertThat(prb4.hasIdentifier()).isTrue();
     assertThat(prb4.getIdentifier()).hasSize(2);
     assertThat(identifier1Value).isEqualTo("78654");
-    assertThat(identifier1System).isNull();
+    assertThat(identifier1System).isEqualTo("urn:id:78654");
```
jgrant22, line 229: PV1.19.4 doesn't give any issues but PID.18.4 is only reading 18.1 even when I add a PID18.4 to the segment

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -295,6 +297,16 @@
...
     };
+    public static final ValueExtractor<Object,String> VISIT_NUMBER_SYSTEM = (Object value) -> {
+
+       HD codingSystem = ((CX) value).getCx4_AssigningAuthority();
+
+      if (codingSystem != null) {
+            String system =  "urn:id:" + codingSystem.getHd1_NamespaceID();
+          return system;
+      }
+        else return null;
```
LisaWellman, line 308: 'else' keyword can be removed

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -295,6 +297,16 @@
         }
         else return text;
     };
+    public static final ValueExtractor<Object,String> VISIT_NUMBER_SYSTEM = (Object value) -> {
```
LisaWellman, line 300: please add a method comment that describes this and fix vertical spacing

src/main/java/io/github/linuxforhealth/hl7/message/util/SegmentExtractorUtil.java
```diff
@@ -247,9 +247,12 @@ private SegmentExtractorUtil() {}
       values = getChildStructures(commonParent, relativeGroupsToCommonParent, seg.getSegment(),
           dataExtractor);
 
-    } else {
-
-      throw new IllegalStateException("Unknow state , cannot extract additional elements");
+    } else  {
```
LisaWellman, line 250: Add a code comment when this leg is hit (and add other comments about anything you discovered when debugging this that would help future coders)

src/main/resources/hl7/message/PPR_PC1.yml
```diff
@@ -48,6 +48,7 @@ resources:
       repeats: true
       additionalSegments:
         - MSH
+        - PATIENT_VISIT.PV1
```
LisaWellman, line 51: Doesn't this have to be PATIENT.PATIENT_VISIT.PV1 ?

src/main/resources/hl7/datatype/Identifier_var.yml
```diff
@@ -10,11 +10,17 @@ type:
...
-system:
+system_1:
    type: STRING
    valueOf: $systemId 
 
+system_2:
+   condition: $systemCX NOT_NULL
+   type: STRING
+   valueOf: $join
+   vars:
```
LisaWellman, line 25: Can you put `sys: "urn:id:"` here rather than in all the individual resource ymls?

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7IdentifierFHIRConversionTest.java
```diff
@@ -612,7 +671,7 @@ public void medicationRequest_identifier_test() {
 
     assertThat(medReq.hasIdentifier()).isTrue();
     assertThat(medReq.getIdentifier()).hasSize(2);
-    assertThat(medReqValue).isEqualTo("20170215080000");
+    assertThat(medReqValue).isEqualTo("8846511");
```
LisaWellman, line 674: This is a great fix that shows the change is working, we are now getting `8846511 `from **PV1-19**.
However, the previous value `20170215080000 `was coming from **MSH-7**.
This message also has a value in **PID-18**, `78654`.  I am concerned that it was never working correctly to pick up PV1-19 **or** PID-18.  Please add a test where is should pick up PID-18.

LisaWellman (commented):
Tough review, it is hard to see the forest at this point.
I'd like to take another shot at looking what we are testing and maybe do some cleanup and add any missing tests, but that can be after this PR.

One main comment about adding a test that verifies PID-18 for patient visit

src/main/resources/hl7/message/PPR_PC1.yml
LisaWellman: This same change will be needed in PPR_PC2.yml and PPR_PC3.yml

LisaWellman:
Don't we need to add 
   additionalSegments: PID
To all the messages that have Encounter?
Do we have a test that is using PID-18 in an Encounter?

src/main/resources/hl7/message/RDE_O11.yml
```diff
@@ -57,5 +57,6 @@ resources:
     resourcePath: resource/MedicationRequest
     repeats: true
     additionalSegments:
+      - PATIENT.PATIENT_VISIT.PV1
```
LisaWellman, line 60: Do we also need PID as an additional segment?  Do we have a test for using PID-18 in a MedicationRequest?

LisaWellman:
Don't we need to add
`additionalSegments: PID, PV1`
To all the messages that have Procedure ?
Do we have a test that is using PID-18, PV1-19 in an Procedure ?

Same for ServiceRequest.

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
jgrant22: I can actually get rid of this with the new method

src/main/resources/hl7/message/PPR_PC1.yml
jgrant22: no ppr_pc1 doesn't have a patient group. Just a patient_visit and an order group

src/main/resources/hl7/datatype/Identifier_var.yml
jgrant22: yep!

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7IdentifierFHIRConversionTest.java
jgrant22: PID-18 test added

src/main/resources/hl7/message/RDE_O11.yml
jgrant22: yep thanks!

jgrant22:
> Don't we need to add
> `additionalSegments: PID, PV1`
> To all the messages that have Procedure ?
> Do we have a test that is using PID-18, PV1-19 in an Procedure ?
> 
> Same for ServiceRequest.

added! thanks

jgrant22:
> Don't we need to add
> `additionalSegments: PID, PV1`
> To all the messages that have Procedure ?
> Do we have a test that is using PID-18, PV1-19 in an Procedure ?
> 
> Same for ServiceRequest.

added! thanks

jgrant22:
> Don't we need to add
> `additionalSegments: PID`
> To all the messages that have Encounter?
> Do we have a test that is using PID-18 in an Encounter?

added thanks!

LisaWellman (approved):

PR #170, opened 2021-08-20
620 config home env

evbaron:
Introduced HL7CONVERTER_CONFIG_HOME environment that should point to the location (folder) of the config.properties. 
The order of operations is to look for Config Directory and find first non empty entry:
* Environment variable
* System Properties

If Config Directory is empty the Classpath Location Strategy is used to search local resource folder

### DIFF
--- a/src/main/java/io/github/linuxforhealth/core/config/ConfigDirectoryLocationStrategy.java
+++ b/src/main/java/io/github/linuxforhealth/core/config/ConfigDirectoryLocationStrategy.java
@@ -17,8 +17,7 @@ import org.slf4j.LoggerFactory;
 
 public class ConfigDirectoryLocationStrategy implements FileLocationStrategy {
 
-  private static final String ENV_CONF_PROP_HOME = "hl7converter.config.home";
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/linuxforhealth/core/config/ConfigDirectoryLocationStrategy.java
```diff
@@ -17,7 +17,7 @@
 
 public class ConfigDirectoryLocationStrategy implements FileLocationStrategy {
 
-  private static final String CONF_PROP_HOME = "config.home";
+  private static final String CONF_PROP_HOME = "HL7CONVERTER_CONFIG_HOME";
```
LisaWellman, line 20: I was expecting a similar environment variable and property name, but keeping to standard naming convention for each, eg:
```
private static final String CONF_PROP_HOME = "hl7converter.config.home"
private static final String ENV_VAR_HOME = "HL7CONVERTER_CONFIG_HOME"
```

LisaWellman (approved):

PR #182, opened 2021-08-25
Final additions + polish

xpander53:


cragun47 (commented):
I agree removing `vars: code: PRB.13` and other related because they don't add anything to the CWE calculation.

PR #192, opened 2021-08-30
onsetfix

xpander53:


src/main/resources/hl7/resource/Condition.yml
```diff
@@ -154,12 +154,12 @@ onsetDateTime:
   expressionType: HL7Spec
   
 onsetString:
-  condition: $onsetDateTime NOT_NULL
+  condition: $onsetDateTime NULL
   type: STRING
   valueOf: PRB.17
```
klwhaley, line 159: concerned about value of... being only PRB.17...shoudln't there be a DG1 value if we use DG1.5?

klwhaley: After discussing, remove DG1.5 and add comment:
# DG1.5 not referenced for onsetDateTime as there is currently not a fallback DG1 segment to use for a onset string

klwhaley (approved):
Approving with comment

src/main/resources/hl7/resource/Condition.yml
xpander53: Removing DG1.5 and added a comment.

PR #197, opened 2021-08-31
practitioner update

xpander53:


klwhaley (approved):
Also reviewed additional code that went directly to master

PR #199, opened 2021-09-01
"ICD-10" missing from CodingSystemMappings

cragun47:

We had ICD10 and I10, but not ICD-10 as a mapping.  Simple change and small test addition to catch it.


klwhaley (approved):
Looks good!

PR #205, opened 2021-09-03
fix performer actor

xpander53:


src/main/resources/hl7/secondary/Performer.yml
```diff
@@ -21,11 +21,8 @@ function_v2:
     code: AP
     text: Administering Provider
 actor:
-   condition: $actorVal NOT_NULL
```
klwhaley, line 24: I don't think the fall back works in vars. 

update the condition to be:
condition: $orderingProvider || $administeringProvider

Can we also add tests for this? When one of those exists, and when neither exist (ie no performer is created).

xpander53: That didn't work but I got it to work adding NOT_NULL after each variable.
I added one test for this - if you want to add more then let's talk about what we are looking for.

src/test/java/io/github/linuxforhealth/MedicationFHIRConverterTest.java
```diff
@@ -19,145 +20,173 @@
...
+        if (messageHeaderExpected) {
+            List<Resource> messageHeader = e.stream()
+                    .filter(v -> ResourceType.MessageHeader == v.getResource().getResourceType())
+                    .map(BundleEntryComponent::getResource).collect(Collectors.toList());
+            assertThat(messageHeader).hasSize(1);
+        }
+
+    }
+
+    // Tests that onset[x] is correctly set to PRB.17 if we have no PRB.16
```
klwhaley, line 150: Test looks good! can you fix this comment to be about this test.

klwhaley (approved):
Approving with nit on comment

src/test/java/io/github/linuxforhealth/MedicationFHIRConverterTest.java
xpander53: Fixed

PR #213, opened 2021-09-09
Issue#212 provide text when unknown id codings

cragun47:

Code adjusted so that IDs that look up to a valid coding and system, the coding and system are provided, but not text.  FHIR search will still work because the the original code value is in the coding.  Text is unneeded, and is duplicate.

When a code does not resolve to the implied system, the type will contain only a text field with the code.

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
@@ -324,6 +324,25 @@ public class SimpleDataValueResolver {
         return value;
     };
 
+    // Special case of a SYSTEM V2.  Identifiers allow unkown codes.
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -324,6 +324,25 @@
         return value;
     };
 
+    // Special case of a SYSTEM V2.  Identifiers allow unkown codes.
```
klwhaley, line 327: nit: unknown misspelled

cragun47: Fixed

klwhaley (approved):

PR #221, opened 2021-09-13
Issue#220 additional obx support

cragun47:
Add support for:

Category=laboratory when SPM is present
Identifier from OBX.21 when present.
Add purpose to Organization Contact.

Fixes exceptions related to NTE's.

### DIFF
--- a/src/main/resources/hl7/resource/Immunization.yml
+++ b/src/main/resources/hl7/resource/Immunization.yml
@@ -100,17 +100,7 @@ note:
    vars:
       obx3: STRING, OBX.3.1
       text: STRING, OBX.5
-reasonCode:
-   valueOf: datatype/CodeableConcept_var
-   expressionType: resource
-   generateList: true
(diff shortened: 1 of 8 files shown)


src/main/resources/hl7/resource/Immunization.yml
```diff
@@ -100,17 +100,7 @@ note:
    vars:
       obx3: STRING, OBX.3.1
       text: STRING, OBX.5
-reasonCode:
```
LisaWellman, line 103: Not clear what drove this change, can you explain?  Is this now covered elsewhere?

LisaWellman (commented):
One concern regarding Immunization

src/main/resources/hl7/resource/Immunization.yml
LisaWellman: I see it was an issue noted long ago from VXU messages

LisaWellman (approved):

PR #228, opened 2021-09-15
encounter.diagnosis fix

xpander53:


src/test/java/io/github/linuxforhealth/hl7/segments/HL7ConditionFHIRConversionTest.java
```diff
@@ -258,9 +260,85 @@ public void validateEncounterMultipleDiagnoses() {
```
klwhaley, line 260: Please add a check like this:

        // Verify reasonReference has a reference for each condition.
        List<Base> reasonReferences = encounter.getNamedProperty("reasonReference").getValues();
        // Therefore there should be 1.
        assertThat(reasonReferences).hasSize(1);

src/test/java/io/github/linuxforhealth/hl7/segments/HL7ConditionFHIRConversionTest.java
```diff
@@ -248,7 +249,8 @@ public void validateEncounterMultipleDiagnoses() {
 
         Coding condCoding = condCC.getCoding().get(0);
         assertThat(condCoding.hasSystem()).isTrue();
-        // change from http://hl7.org/fhir/sid/icd-10 to http://hl7.org/fhir/sid/icd-10-cm temporarily, see Issue #189
+        // change from http://hl7.org/fhir/sid/icd-10 to
+        // http://hl7.org/fhir/sid/icd-10-cm temporarily, see Issue #189
         assertThat(condCoding.getSystem()).isEqualTo("http://hl7.org/fhir/sid/icd-10-cm");
```
klwhaley, line 254: Can we also a check that condition.encounter is set correctly?

src/test/java/io/github/linuxforhealth/hl7/segments/HL7ConditionFHIRConversionTest.java
xpander53: Added

src/test/java/io/github/linuxforhealth/hl7/segments/HL7ConditionFHIRConversionTest.java
xpander53: added

PR #229, opened 2021-09-16
adt-a40 messages to use adt-a39 structure

lisadier:
Signed-off-by: lisadier <lisadier@us.ibm.com>

adt-a40 yaml has been updated to reflect the default structure of an adt-a40 message: https://hl7-definition.caristix.com/v2/HL7v2.6/TriggerEvents/ADT_A40

The default structure of adt-a40 is same as adt-a39.

The unit tests have been updated accordingly.


.gitignore
```diff
@@ -45,6 +45,10 @@ bin/
 # Jetbrains IDEA
 .idea
 
+# VSCode
+.vscode/settings.json
+
 # Local History files
 .history
 AnnaConverterFolderTest.java
```
LisaWellman, line 53: We should not have these Anna* files in listed here in the repo

lisadier: Removed both the one that was there already and the one I added.  We'll create these as a tool, in a better way, in a future work item.

LisaWellman (approved):
It would be nice to add a TODO to the disabled tests indicating they should be enabled when these msg events are supported (with the MRG work)

PR #235, opened 2021-09-17
YAML update for groups

lisadier:
Signed-off-by: Lisa Dierkhising <lisadier@us.ibm.com>

ADT_A01: 
In MessageHeader, removed MSH from additionalSegments because it is the segment, not needed in both.
In Observation, ORC segment is not in ADT_A01(not even the sub-groups), so removed it.
In Procedure, ORC and OBR segments are not in ADT_A01(not even the sub-groups), so removed it.

MDM_T02:
In Observation, OBX is hidden within OBSERVATION (even though caristix calls the group ‘OBXNTE’, using the debugger I found that the parser encodes it in group ‘OBSERVATION’, also confirmed with HL7 spec)
In DocumentReference, ORC and OBR are within COMMON_ORDER. ***TODO: I haven’t yet figured out how to get around this exception: Additional segements cannot be from a group if primary segment is not defined to be from same group. I decided to deliver because this is a better state than it was in, but I'll have to look at this more on Monday.

MDM_T06:
In Observation, OBX is hidden within OBSERVATION (even though Caristix calls it OBXNTE).

MDM_T04, T08, and T10:
In Observation, OBX is hidden within OBSERVATION (even though Caristix calls it OBXNTE). ***Note: couldn’t find a test to verify my changes were ok.

OMP_O09:
AL1 is within the PATIENT group.

ORM_O01:
There is no EVN in Caristix nor the HL7 spec, so removed it from MessageHeader
OBX is nested in ORDER.ORDER_DETAIL.OBSERVATION.OBX
AL1 is within PATIENT
DG1 is within ORDER.ORDER_DETAIL

OUR_R01:
In Patient resource, I don’t think that PD1 will get picked up because there is no group defined.  —I tried to fix but still not getting the Patient.generalPractitioner set.  (I've started a new test test_oru_patient_general_practitioner which I'll deliver in a future PR if we decide we want to have a test for this.)

PPR_PC1, PPR_PC2, PPR_PC3:
Removed PD1 from Patient because that segment isn’t available in this message (confirmed on Caristix and in our MessageTypesToSegments ss).

### DIFF
--- a/src/main/resources/hl7/message/RDE_O11.yml
+++ b/src/main/resources/hl7/message/RDE_O11.yml
@@ -34,7 +34,7 @@ resources:
       - MSH
       
   - resourceName: AllergyIntolerance
-    segment: AL1
+    segment: PATIENT.AL1
     resourcePath: resource/AllergyIntolerance
     repeats: true
     additionalSegments:
--- a/src/main/resources/hl7/message/RDE_O25.yml
+++ b/src/main/resources/hl7/message/RDE_O25.yml
@@ -45,7 +45,7 @@ resources:
       - PATIENT.PID
 
   - resourceName: AllergyIntolerance
-    segment: AL1
+    segment: PATIENT.AL1
     resourcePath: resource/AllergyIntolerance
     repeats: true
     additionalSegments:
--- a/src/main/resources/hl7/message/ADT_A01.yml
+++ b/src/main/resources/hl7/message/ADT_A01.yml
@@ -13,7 +13,6 @@ resources:
       isReferenced: false
       additionalSegments: 
              - EVN
-             - MSH
 
     - resourceName: Patient
       segment: PID
@@ -52,7 +51,6 @@ resources:
       repeats: true
       isReferenced: true
       additionalSegments:
-        - ORC
         - MSH
 
     - resourceName: AllergyIntolerance
@@ -72,5 +70,3 @@ resources:
         - MSH
         - PID
         - PV1
-        - ORC
-        - OBR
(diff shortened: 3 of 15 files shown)


src/main/resources/hl7/message/RDE_O11.yml
```diff
@@ -34,7 +34,7 @@ resources:
       - MSH
       
   - resourceName: AllergyIntolerance
-    segment: AL1
+    segment: PATIENT.AL1
```
LisaWellman, line 37: Should we have a GROUP?

src/main/resources/hl7/message/RDE_O25.yml
```diff
@@ -45,7 +45,7 @@ resources:
       - PATIENT.PID
 
   - resourceName: AllergyIntolerance
-    segment: AL1
+    segment: PATIENT.AL1
```
LisaWellman, line 48: Should there be a GROUP?

LisaWellman (approved):
You mentioned "See new test test_oru_patient_general_practitioner" in your comments, did you mean to deliver that or is it in next PR?
Approving with just one question on the RDE's allergy

src/main/resources/hl7/message/RDE_O11.yml
lisadier: This is actually a valid way to do it without a group specified.  Since only this one field is used from the Patient group it is a simpler syntax.

src/main/resources/hl7/message/RDE_O25.yml
lisadier: This is actually a valid way to do it without a group specified. Since only this one field is used from the Patient group it is a simpler syntax.

lisadier:
Good point, I'll update my description.  Since I didn't get it working, I didn't deliver the test.  We should probably talk about whether I've found a real bug around GeneralPractitioner or if it just isn't implemented yet.

PR #237, opened 2021-09-20
Obx sn support

klwhaley:
1. SN type support for OBX
2. Encounters referencing OBX's

### DIFF
--- a/src/main/resources/hl7/datatype/Quantity.yml
+++ b/src/main/resources/hl7/datatype/Quantity.yml
@@ -10,6 +10,7 @@ value :
    valueOf: $value
 comparator : 
    type: String
+   condition: $comparator EQUALS '<' || $comparator EQUALS '<=' || $comparator EQUALS '>=' || $comparator EQUALS '>'
    valueOf: $comparator
 unit : 
    type: String
--- a/src/main/resources/hl7/datatype/Quantity.yml
+++ b/src/main/resources/hl7/datatype/Quantity.yml
@@ -10,6 +10,7 @@ value :
    valueOf: $value
 comparator : 
    type: String
+   # Hl7 supports an equal sign as a comparator that FHIR does not, making sure its a supported comparator.
    condition: $comparator EQUALS_STRING '<' || $comparator EQUALS_STRING '<=' || $comparator EQUALS_STRING '>=' || $comparator EQUALS_STRING '>'
    valueOf: $comparator
 unit : 
--- a/src/main/resources/hl7/datatype/Quantity.yml
+++ b/src/main/resources/hl7/datatype/Quantity.yml
@@ -10,7 +10,7 @@ value :
    valueOf: $value
 comparator : 
    type: String
-   condition: $comparator EQUALS '<' || $comparator EQUALS '<=' || $comparator EQUALS '>=' || $comparator EQUALS '>'
+   condition: $comparator EQUALS_STRING '<' || $comparator EQUALS_STRING '<=' || $comparator EQUALS_STRING '>=' || $comparator EQUALS_STRING '>'
    valueOf: $comparator
 unit : 
    type: String
@@ -22,5 +22,6 @@ system_2:
    type: STRING
    valueOf: $resolvedSystem   
 code : 
+   condition: $system NOT_NULL || $resolvedSystem NOT_NULL
    type: String
    valueOf: $code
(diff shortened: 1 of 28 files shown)


src/main/resources/hl7/datatype/Quantity.yml
```diff
@@ -10,6 +10,7 @@ value :
    valueOf: $value
 comparator : 
    type: String
+   condition: $comparator EQUALS_STRING '<' || $comparator EQUALS_STRING '<=' || $comparator EQUALS_STRING '>=' || $comparator EQUALS_STRING '>'
```
LisaWellman, line 13: This is everything but equals?  Might add a comment

src/main/resources/hl7/resource/Observation.yml
```diff
@@ -76,16 +76,42 @@ valueString:
...
+    vars: 
+      value: OBX.5.2
+      unit: OBX.6.1 
+      code: OBX.6.1 
+      comparator: OBX.5.1
+      obx2: STRING, OBX.2
+      separator: STRING, OBX.5.3
+      resolvedSystem: SYSTEM_URL, OBX.6.3
+
+valueQuantity_3:
```
LisaWellman, line 103: I am not seeing any difference from valueQuantity_2, am I missing it?

src/main/resources/hl7/resource/Observation.yml
```diff
@@ -120,17 +146,27 @@ valueTime:
...
+   expressionType: resource
+   specs: OBX.5
+   vars:
+     numerator: SN.2
+     denominator: SN.4
+     obx2: STRING, OBX.2
+     separator: STRING, SN.3
+
+valueRatio_2:
+   condition: $obx2 EQUALS SN && $separator NOT_NULL && $separator EQUALS '/'
```
LisaWellman, line 161: This looks the same as value_ratio1, can they be combined?

src/test/java/io/github/linuxforhealth/hl7/message/DifferentObservationValueTest.java
```diff
@@ -6,6 +6,10 @@
 package io.github.linuxforhealth.hl7.message;
```
LisaWellman, line 6: I think we should move this to io.github.linuxforhealth.hl7.segments

I also find the name strange, it probably had a specific purpose but is now more general.  I think it should just be a general tests class for Observation - remove "Different" and maybe also "Value"?

src/test/java/io/github/linuxforhealth/hl7/message/DifferentObservationValueTest.java
```diff
@@ -39,377 +43,571 @@
...
+        Quantity q = obs.getValueQuantity();
+        assertEquals("ug/mL", q.getCode()); //code for unit
+        assertEquals("ug/mL", q.getUnit()); //setting unit to OBX.6.1
+        assertEquals("http://unitsofmeasure.org", q.getSystem());  //system for unit
+        assertEquals(0.06f, q.getValue().floatValue());
+        assertEquals("<", q.getComparator().toCode());
+    }
+
+    /**
+     * Testing Observation.yml valueQuantity_2
```
LisaWellman, line 128: make comment indicate how this is different than previous test

LisaWellman (approved):

src/test/java/io/github/linuxforhealth/hl7/message/DifferentObservationValueTest.java
klwhaley: agreed, considered it already, done.

src/main/resources/hl7/resource/Observation.yml
klwhaley: difference is separator.  can't do complex conditions...otherwise i would have one :

obx2 equals SN && separator not null && (separator equals / || separator equals :)

no change at this time.

src/main/resources/hl7/resource/Observation.yml
klwhaley: quanity_2 is separator is null..quantity 3 is separator isnot null and is not : or / 

again no complex conditions :(

no change at this time.

src/main/resources/hl7/datatype/Quantity.yml
klwhaley: added comment clarifying.

src/test/java/io/github/linuxforhealth/hl7/message/DifferentObservationValueTest.java
klwhaley: fixed

PR #238, opened 2021-09-20
Ensure Encounter refs OBX and DG1

cragun47:
Encounter was not making both references when OBX and DG1 were both present.


klwhaley (approved):
looks good!

PR #239, opened 2021-09-20
A40 MRG fixes

xpander53:
Signed-off-by: Joel Maner <xpander53@yahoo.com>

### DIFF
--- a/src/main/resources/hl7/resource/PatientMRG.yml
+++ b/src/main/resources/hl7/resource/PatientMRG.yml
@@ -36,17 +36,35 @@ active:
    type: BOOLEAN  
    valueOf: false
 
-link:
+link_1:
    generateList: true
    expressionType: nested
-   condition: $Patient NOT_NULL
+   condition: $triggerEvent EQUALS A34
+   vars: 
+      triggerEvent: STRING, MSH.9.2
    expressionsMap:
-         type:
-            type: STRING
-            valueOf: 'replaced-by'
-         other:
-           required: true
-           valueOf: datatype/Reference
-           expressionType: resource
-           specs: $Patient
+      type:
+        type: STRING
+        valueOf: 'replaced-by'
+      other:
+        required: true
+        valueOf: datatype/Reference
+        expressionType: resource
+        specs: $Patient
+link_2:
+   generateList: true
+   expressionType: nested
+   condition: $triggerEvent EQUALS A40
+   vars: 
+      triggerEvent: STRING, MSH.9.2
+   expressionsMap:
+      type:
+        type: STRING
+        valueOf: 'replaced-by'
+      other:
+        required: true
+        valueOf: datatype/Reference
+        expressionType: resource
+        specs: $Patient
+        useGroup: true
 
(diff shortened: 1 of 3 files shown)


lisadier (approved):
Looks great!

src/main/resources/hl7/resource/PatientMRG.yml
```diff
@@ -36,17 +36,35 @@ active:
...
-           specs: $Patient
+      type:
+        type: STRING
+        valueOf: 'replaced-by'
+      other:
+        required: true
+        valueOf: datatype/Reference
+        expressionType: resource
+        specs: $Patient
+link_2:
```
klwhaley, line 54: can we add a comment that link_2 is due to different structure of segment groups in A40, so we remember later why we did this?

src/test/java/io/github/linuxforhealth/hl7/segments/HL7MergeFHIRConversionTest.java
```diff
@@ -159,10 +158,9 @@ public void validateHappyPathADT_A34WithMRG() {
```
klwhaley, line 158: Do we have an ADT_A34 test?

src/test/java/io/github/linuxforhealth/hl7/segments/HL7MergeFHIRConversionTest.java
```diff
@@ -253,6 +250,110 @@ public void validateTwoMRGs() {
...
+
+       // Verify the patient four identifier value
+       assertThat(ResourceUtils.getValueAsString(patientFourIdentifierList.get(0), "value")).isEqualTo("MR4");
+       // Verify the patient four identifier value is marked as old
+       assertThat(ResourceUtils.getValueAsString(patientFourIdentifierList.get(0), "use")).isEqualTo("Enumeration[old]");
+
+       //The fourth patient should NOT be active.
+       assertThat(ResourceUtils.getValueAsString(patientFour, "active")).isEqualTo("BooleanType[false]");
+
+       // We should have link.other.reference to the PID (3r) patient
```
klwhaley, line 351: nit 3r? 3rd?

xpander53: Fixed

src/test/java/io/github/linuxforhealth/hl7/segments/HL7MergeFHIRConversionTest.java
xpander53: Yes it's the first test in the suite

src/main/resources/hl7/resource/PatientMRG.yml
xpander53: Will do

LisaWellman, line 54: looks like this is the same as the previous, can it not be combined (would that be a 'complex condition')?

src/test/java/io/github/linuxforhealth/hl7/segments/HL7MergeFHIRConversionTest.java
klwhaley: just missed it, closing

src/main/resources/hl7/resource/PatientMRG.yml
LisaWellman: It is not clear to me why we need the condition check -- when would we NOT want these to fire?

klwhaley (approved):
approving with a few minor comments

src/main/resources/hl7/resource/PatientMRG.yml
xpander53: I added a comment to explain.
It's triggering different "link"s depending on the event trigger.
A40 has patient in a group and A34 doesn't.

PR #243, opened 2021-09-23
fixes for 345 

jgrant22:
fix participant to show display instead of text, fix enc.class to show just the code when the mappings aren't correct


src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -250,7 +250,7 @@
         } else {
             // defaults to unknown
```
LisaWellman, line 251: fix comment

src/main/resources/hl7/datatype/Duration.yml
```diff
@@ -5,37 +5,24 @@
...
-    vars: 
+    vars:
       var1: $time1
       var2: $time2
 
 value_2:
-  condition: $var3 NOT_NULL && $var1 NULL
-  valueOf: $var3
-  expressionType: HL7Spec
+  condition: $var1 NOT_NULL && $time1 NULL
```
LisaWellman, line 15: can we make this
```
$time3 NOT_NULL && $time1 NULL
valueOf: $time3
```
and remove 
```
vars:
var1: $time3
```
this makes it line up with what was done for unit_2

src/test/java/io/github/linuxforhealth/hl7/segments/HL7EventTypeFHIRConversionTest.java
```diff
@@ -53,11 +53,6 @@ public void validate_evn_segment() {
         .collect(Collectors.toList());
     assertThat(encounterResource).hasSize(1);
 
-    // ENV.4 is used for reasonCode
```
LisaWellman, line 56: Was it verified to prioritize PV2-3 and not EVN-4 for now?
Please add an item to the deferred work in the card to track adding EVN-4 back in.  We should probably comment these lines out, rather than remove, and mark them with a TODO, and then reference in that deferred work to uncomment these.
This might also cause GT changes, so it needs to be communicated to test.

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
```diff
@@ -164,47 +164,6 @@ public static Long diffDateMin(Object start, Object end) {
         return null;
     }
 
-    public static Long getTimeDiff(Object start, Object end) {
```
LisaWellman, line 167: bummer :-( sorry

LisaWellman (approved):
Looks good, approving with a couple small items

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
jgrant22: ooops thanks!

src/main/resources/hl7/datatype/Duration.yml
jgrant22: done

claverols:
I'm not able to merge these approved changes, as I have no write access to this repository. 
@LisaWellman @jgrant22 can you please do so when you get a chance? Thanks

src/test/java/io/github/linuxforhealth/hl7/segments/HL7EventTypeFHIRConversionTest.java
jgrant22: I removed this because of the IS change EVN.4  is the first option still but this test was looking for the text

LisaWellman: Oh, I see -- I think we need to change the test to look at the code, not the text

PR #256, opened 2021-09-29
Fix Encounter phantom length units, better tests

cragun47:

Fixes edge case where we were seeing "phantom" empty lengths with units but no value.
Needed to check for missing PV1.44 or PV1.45 _AND_ missing PV2.11.
Added test.
Minimized input record data in all length tests for clarity.

### DIFF
--- a/src/main/resources/hl7/datatype/Duration.yml
+++ b/src/main/resources/hl7/datatype/Duration.yml
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 ---
+# value_1 is calculated wnen time1 and time2 are present 
 value_1:
     valueOf: 'GeneralUtils.diffDateMin(var1,var2)'
     expressionType: JEXL
@@ -11,13 +12,18 @@ value_1:
       var1: $time1
       var2: $time2
 
+# value_2 or _3 comes from $time3 if either $time1 or $time2 is missing
 value_2:
   condition: $time3 NOT_NULL && $time1 NULL
   valueOf: $time3
 
-
+# (near duplicate of value_2 to handle complex boolean) 
+value_3:
+  condition: $time3 NOT_NULL && $time2 NULL
+  valueOf: $time3  
+  
 unit_1:
-  condition: $time1 NOT_NULL
+  condition: $time1 NOT_NULL && $time2 NOT_NULL
   type: String
   value:  "Minutes"
 
--- a/src/main/resources/hl7/datatype/Duration.yml
+++ b/src/main/resources/hl7/datatype/Duration.yml
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 ---
-# value_1 is calculated wnen time1 and time2 are present 
+# value_1 is calculated when time1 and time2 are present 
 value_1:
     valueOf: 'GeneralUtils.diffDateMin(var1,var2)'
     expressionType: JEXL
--- a/src/main/resources/hl7/datatype/Duration.yml
+++ b/src/main/resources/hl7/datatype/Duration.yml
@@ -25,9 +25,31 @@ value_3:
 unit_1:
   condition: $time1 NOT_NULL && $time2 NOT_NULL
   type: String
-  value:  "Minutes"
+  value:  "minutes"
 
 unit_2:
   condition: $time3 NOT_NULL
   type: String
-  valueOf: "Days"
+  valueOf: "days"
+
+code_1:
+  condition: $time1 NOT_NULL && $time2 NOT_NULL
+  type: String
+  value:  "min"
+
+code_2:
+  condition: $time3 NOT_NULL
+  type: String
+  valueOf: "d"  
+
+# system is hardcoded to UCUM, but must only show if other values show
+system_1:
+  condition: $time1 NOT_NULL && $time2 NOT_NULL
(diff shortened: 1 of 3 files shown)


src/main/resources/hl7/datatype/Duration.yml
```diff
@@ -4,20 +4,26 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 ---
+# value_1 is calculated wnen time1 and time2 are present
```
LisaWellman, line 7: Thank you for adding the comments!

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7EncounterFHIRConversionTest.java
```diff
@@ -424,12 +424,14 @@ public void test_encounter_reason_code() {
...
 
     @Test
-    public void test_encounter_length() {
-        //When length between encounters is a day or more apart the units should be "Days"
-        String hl7message = "MSH|^~\\&|PROSOLV|SENTARA|WHIA|IBM|20151008111200|S1|ADT^A01^ADT_A01|MSGID000001|T|2.6|10092|PRPA008|AL|AL|100|8859/1|ENGLISH|ARM|ARM5007\n"
-                + "EVN|A04|20151008111200|20171013152901|O|OID1006|20171013153621|EVN1009\n"
+    public void testEncounterLength() {
+        // When length between encounters is a day or more apart the units should be "Days"
+        // Both start PV1.44 and end PV1.45 must be present to use either value as part of lengthh
+        String hl7message = "MSH|^~\\&|PROSOLV||||20151008111200||ADT^A01^ADT_A01|MSGID000001|T|2.6|||||||||\n"
```
LisaWellman, line 430: Thank you for simplifying the messages, this makes it so much easier to see the relevant fields!

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7EncounterFHIRConversionTest.java
```diff
@@ -438,11 +440,13 @@ public void test_encounter_length() {
         assertThat(encounterLength.getValue()).isEqualTo((BigDecimal.valueOf(1440)));
         assertThat(encounterLength.getUnit()).isEqualTo("Minutes");
 
-        //When length between encounters is a less than a apart the units should be "Minutes"
-        hl7message = "MSH|^~\\&|PROSOLV|SENTARA|WHIA|IBM|20151008111200|S1|ADT^A01^ADT_A01|MSGID000001|T|2.6|10092|PRPA008|AL|AL|100|8859/1|ENGLISH|ARM|ARM5007\n"
-                + "EVN|A04|20151008111200|20171013152901|O|OID1006|20171013153621|EVN1009\n"
+        // When length between encounters is a less than a apart the units should be "Minutes"
```
LisaWellman, line 443: Can you fix this comment, it does not read correctly.  At one point we were deciding between days and minutes depending on the duration, but I do not think that is happening anymore so maybe it can be deleted.

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7EncounterFHIRConversionTest.java
```diff
@@ -451,24 +455,39 @@ public void test_encounter_length() {
...
         encounterLength = encounter.getLength();
         assertThat(encounterLength.getValue()).isEqualTo((BigDecimal.valueOf(3)));
-        assertThat(encounterLength.getUnit()).isEqualTo("Days");
+        assertThat(encounterLength.getUnit()).isEqualTo("Days");  
+
+        // If PV1.44 or PV1.45 are missing, and there is no PV2.11 for back-up, no length is created
+        hl7message = "MSH|^~\\&|PROSOLV||||20151008111200||ADT^A01^ADT_A01|MSGID000001|T|2.6|||||||||\n"
+                + "EVN|A04|20151008111200|||||\n"
+                + "PID|||1234^^^^MR||DOE^JANE^|||F||||||||||||||||||||||\n"
+                // PV1.44 empty; PV1.45 present
```
LisaWellman, line 478: This seems to be the same test as above, I assume you meant to change this to be
PV1.44 present; PV1.45 empty

LisaWellman (commented):
Thank you for the great comments and tests.
Do need to update the one test so we are checking both combinations of 44/45 being provided/empty

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7EncounterFHIRConversionTest.java
cragun47: As discussed, we always use minutes.   
I removed one test because a short minutes caculation and a long minutes calculation are now just a repeat of the same, and parsing is expensive (and tests are already slow)

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7EncounterFHIRConversionTest.java
cragun47: Not the same test.  One has PV2.11 and second does not.   
The first proves that PV2 is use if PV1.44/45 is not there.
The second proves if neither is complete, we have no value, which is the problem Anna saw.

LisaWellman (approved):

PR #257, opened 2021-09-29
OML_O21 message structure added

hajiratasneem:
I have added the OML_O21 message structure. But I am unable to generate multiple DiagnosticReport resources in the FHIR bundle that is getting generated even though I provided the 'repeats' key to be True for the ORC and OBR segments, if they are repeated more than once.

PR #271, opened 2021-10-05
fixed failure for OML test

pbhallam:
Signed-off-by: Pushpa Bhallamudi <pbhallam@us.ibm.com>

### DIFF
--- a/src/main/resources/config.properties
+++ b/src/main/resources/config.properties
@@ -1,4 +1,4 @@
 base.path.resource=
-supported.hl7.messages=ADT_A01, ADT_A34, ADT_A40, ORU_R01, PPR_PC1, VXU_V04, RDE_O11, OMP_O09, RDE_O25, MDM_T02
+supported.hl7.messages=ADT_A01, ADT_A34, ADT_A40, ORU_R01, PPR_PC1, VXU_V04, RDE_O11, OMP_O09, RDE_O25, MDM_T02,OML_O21
 default.zoneid=+08:00
 additional.conceptmap.file=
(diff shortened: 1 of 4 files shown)


pbhallam:
@hajiratasneem  I have updated your OML message YAML and modified to make it work with HAPI parser non-greedy mode.

Here are details about that: 
https://hapifhir.github.io/hapi-hl7v2/base/apidocs/index.html?ca/uhn/hl7v2/parser/ParserConfiguration.html


public void setNonGreedyMode(boolean theNonGreedyMode)
If set to true (default is false), pipe parser will be put in non-greedy mode. This setting applies only to Pipe Parsers and will have no effect on XML Parsers.
In non-greedy mode, if the message structure being parsed has an ambiguous choice of where to put a segment because there is a segment matching the current segment name in both a later position in the message, and in an earlier position as a part of a repeating group, the earlier position will be chosen.

This is perhaps best explained with an example. Consider the following structure:

 MSH
 GROUP_1 (start)
 {
    AAA
    BBB
    GROUP_2 (start)
    {
       AAA
    }
    GROUP_2 (end)
 }
 GROUP_1 (end)
 
For the above example, consider a message containing the following segments:
MSH
AAA
BBB
AAA
In this example, when the second AAA segment is encountered, there are two possible choices. It would be placed in GROUP_2, or it could be placed in a second repetition of GROUP_1. By default it will be placed in GROUP_2, but in non-greedy mode it will be put in a new repetition of GROUP_1.

This mode is useful for example when parsing OML^O21 messages containing multiple orders.

pbhallam:
@LisaWellman  ^^ fyi

lisadier:
Closing this as this has been resolved in the following PRs:
https://github.com/LinuxForHealth/hl7v2-fhir-converter/pull/257
https://github.com/LinuxForHealth/hl7v2-fhir-converter/pull/413

PR #295, opened 2021-10-13
ORU_R01 hardening

lisadier:
Signed-off-by: Lisa Dierkhising <lisadier@us.ibm.com>

Changes made in this PR:

- Removed unused OBX from ORU_R01 Encounter resource.
- Tests of content for DiagnosticReport
- Additional tests and beefing up of tests to cover most scenarios.  
  - Left to cover in the next PR: the one disabled test case needs more debug and a fix, Multiple OBR with SPM, SPM with OBX, Check actual text in the Observation to confirm right observations
- Updates for including a service request when there is an OBR but no ORC.


lisadier:
I've discussed with Brian.  Rather than holding this PR open, we will handle the changes for his PR in his PR.  So I am going to deliver this, then when Brian is ready, I will help with merging and test updates to this file to handle his changes.

PR #305, opened 2021-10-18
target directory functionality

xpander53:
Signed-off-by: Joel Maner <xpander53@yahoo.com>


dparash:
The PR looks good except for one concern. LOGGER.debug statement spanning more than 1 line will have to be handled appropriately.

PR #322, opened 2021-10-29
Catch and stifle hapi exceptions (phi reasons)

xpander53:


src/main/java/io/github/linuxforhealth/hl7/message/HL7MessageModel.java
```diff
@@ -70,10 +73,18 @@ public String convert(Message message, MessageEngine engine) {
...
-        .encodeResourceToString(bundle);
+    String result = null;
+
+    //Catch any exceptions and don't display them because they *could* show PHI.
+    try {
+        Bundle bundle = engine.transform(dataSource, this.getResources(), new HashMap<>());
+        result = engine.getFHIRContext().encodeResourceToString(bundle);
+    }
+    catch(Exception e) {
+        LOGGER.error("Error transforming HL7 message");
```
LisaWellman, line 84: Thinking about what might be useful here.  I know we saw PHI in the message text, so that's out.  Would the exception type (class) add any value?  I will leave it up to you.

LisaWellman: Might want to add a comment that we have seen PHI in the message text, so someone does not add the message text in the future.

LisaWellman (approved):

src/main/java/io/github/linuxforhealth/hl7/message/HL7MessageModel.java
xpander53: I was thinking the same thing last night - that maybe we could/should include the exception class.

PR #325, opened 2021-10-29
fix issue with hardcoded display showing in meta

jgrant22:
I tried to spend time figuring out how this was happening. I am still not 100% I tried a bunch of things in the yaml just to see if I could isolate my change from the meta stuff but didn't have any luck. I did have luck doing a simple flag that always returns a null display for meta.


LisaWellman:
Seems like we are getting cross-contamination between different CodeableConcepts.  This is scary, I am surprised we have not run into this before.  Searching for CodeableConcept_var there looks like there may be several places where we are not setting display, and so there could be bleed-through.  Adding additional reviews for their thoughts.

Writing this out to get it straight in my head.
1. ExtensionMeta uses CodeableConcept_var, and it in turn uses Coding which uses $display.  Which is of course set by MedicationRequest.yml in requester_1.
2. Practitioner.yml uses Identifier_SystemID, which also uses CodeableConcept_var (which is why you set $display originally).

Making sure there is no display for these meta fields seems reasonable.
Wondering if there is a way to not require the NO_DISPLAY function.  Does it work to do something like `noDisplay: ""`

cragun47 (commented):
I know Jared worked at this a long time.  The NO_DISPLAY $noDisplay "no display" seems really hacky and convoluted.  (No offense, Jared).  If we must use some NO_DISPLAY control, I think I'd rather see a condition in Coding that display_1 $noDisplay NOT_NULL or such. 

Another thing we could consider is some global value $EMPTY or $NULL that could be passed to display (or anything else) indicating that it should be left empty.  I thought NULL would do that.   We have some examples of global variable constants, but they are usually the _presence_ of a value, not the absence of one.

Variables get set and are carried on the variable map for a period, but I don't know if that's the segment or the resource or what.   I don't know when the map gets reset.  

I agree there may be datableed especially if we

src/main/resources/hl7/extension/ExtensionMeta.yml
```diff
@@ -51,8 +51,10 @@ extension_5:
          vars:
             system: SYSTEM_URL, $system_code
             code: String, MSH.9.1
+            display: $NULL
          constants:
             system_code: 'source-record-type-system'
+            noDisplay: "no display"
```
LisaWellman, line 57: This is no longer needed

src/main/resources/hl7/extension/ExtensionMeta.yml
```diff
@@ -67,5 +69,7 @@ extension_6:
          vars:
             system: SYSTEM_URL, $system_code
             code: String, MSH.9.2
+            display: $NULL
          constants:
-            system_code: 'source-event-trigger-system'
\ No newline at end of file
+            system_code: 'source-event-trigger-system'
+            noDisplay: "no display"
```
LisaWellman, line 75: This is no longer needed

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -556,6 +556,10 @@ private static final SimpleCode commonCodingSystemV2 (String table, String code,
         }
     }
 
+    public static final ValueExtractor<Object, String> NO_DISPLAY = (Object value) -> {
```
LisaWellman, line 559: This is no longer needed

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7MedicationRequestFHIRConversionTest.java
```diff
@@ -669,7 +669,7 @@ public void test_MedicationRequest_category_requester_and_dispenseRequest(){
 
         HL7ToFHIRConverter ftv = new HL7ToFHIRConverter();
         String json = ftv.convert(hl7message, PatientUtils.OPTIONS);
-        assertThat(json).isNotBlank();
+        assertThat(json).isNotBlank(); System.out.println(json);
```
LisaWellman, line 672: Remove println

LisaWellman (commented):
Please add a testcase for the bug scenario.

src/main/resources/hl7/extension/ExtensionMeta.yml
jgrant22: oops thanks

jgrant22: done

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7MedicationRequestFHIRConversionTest.java
jgrant22: done

LisaWellman (approved):

PR #326, opened 2021-10-30
Origin/no group 

pbhallam:


cragun47:
@pbhallam A review of the changes looks good so far.  But something is not quite right.  I have added a test case which should work, but fails.  Find `Hl7MDMMessageTest.test_MDMT02_serviceRequest_reference`.  Near the bottom, you will see test code which is commented out.  **_When uncommented, this test code should work_**.

We expect the `$ServiceRequest?` in DocumentReference.yml
```yaml
context:
   valueOf: secondary/Context
   expressionType: resource
   vars:
      timestamp: TXA.4 
      providerCode: TXA.5
      serviceRequestRef: $ServiceRequest?
      encounterRef: $Encounter
```
which is passed to Context.yml
```yaml
related_2:
  valueOf: datatype/Reference
  expressionType: resource
  generateList: true
  specs: $serviceRequestRef
```      

To create the following ServiceRequest reference in context:
```json
"context": {
        "encounter": [ {
                "reference": "Encounter/<GUID>"
        } ],
        "period": {
                "start": "2017-09-20T14:12:33+08:00"
        },
        "related": [ {
                "reference": "Practitioner/<GUID>",
                "display": "JAMES A MATTHEWS"
        }, {
                "reference": "ServiceRequest/<GUID>"
        } ]
        }
```

pbhallam:
@cragun47 -- the issue was with the mdm 02 resource template, if you want to reference service request then you need to ass isreferenced: true

cragun47 (approved):
Looks good. No further concerns.

cragun47:
> @cragun47 -- the issue was with the mdm 02 resource template, if you want to reference service request then you need to ass isreferenced: true

I tried that and you're right.  That was the only concern I has with the code, so I approved.

PR #329, opened 2021-11-01
Add missing PPR_PC1 Order Observations

cragun47:

This solves the problem of missing Observations for PPR_PC1 Order OBX's.
It does not yet solve the issue of the unneeded DocumentationReference.
Hl7PPRMessageTest has been updated for the expected counts.  Comments are also updated to match current capability.

Used opportunity to refactor and streamline Hl7PPRMessageTest using common ResourceUtils methods.  Sorry for the noise.

### DIFF
--- a/src/main/resources/hl7/message/PPR_PC1.yml
+++ b/src/main/resources/hl7/message/PPR_PC1.yml
@@ -58,49 +58,49 @@ resources:
       - .NTE
 
   - resourceName: ServiceRequest
-    segment: .ORDER.ORC
-    group: PROBLEM
+    segment: .ORC
+    group: PROBLEM.ORDER
     resourcePath: resource/ServiceRequest
     repeats: true
     isReferenced: true
     additionalSegments:
-      - .ORDER.ORDER_DETAIL.OBR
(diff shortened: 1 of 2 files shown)


src/main/resources/hl7/message/PPR_PC1.yml
```diff
@@ -69,6 +70,19 @@ resources:
       - MSH
       - PID
 
+  # This OBX is for the Order Observation
+  - resourceName: Observation
+    segment: .ORDER.ORDER_DETAIL.ORDER_OBSERVATION.OBX
+    group: PROBLEM
```
LisaWellman, line 76: Let's make the group `PROBLEM.ORDER` to make it more discernable from line 39, and then the other lines can be simplified since `.ORDER` can be removed

LisaWellman (approved):

src/main/resources/hl7/message/PPR_PC1.yml
cragun47: Will do

cragun47: Done.  Changed other related resource similarly.

PR #330, opened 2021-11-02
Message hardening

lisadier:
Updates made to clean up FHIRConverterTest, and other hardening to remove disabled tests and some TODOs that could be updated, as well as additional coverage.

### DIFF
--- a/src/main/resources/hl7/message/ADT_A08.yml
+++ b/src/main/resources/hl7/message/ADT_A08.yml
@@ -12,7 +12,7 @@ resources:
       repeats: false
       isReferenced: false
       additionalSegments: 
-             - EVN
+        - EVN
    
     - resourceName: Patient
       segment: PID
@@ -20,8 +20,8 @@ resources:
       repeats: false
       isReferenced: true
       additionalSegments:
-             - PD1
-             - MSH
+        - PD1
+        - MSH
 
     - resourceName: Encounter
       segment: PV1
@@ -29,9 +29,10 @@ resources:
       repeats: false
       isReferenced: true
       additionalSegments:
-             - PV2
-             - EVN
-             - MSH
+        - PV2
+        - EVN
+        - MSH
+        - DG1
 
     - resourceName: Condition
       segment: DG1
@@ -39,9 +40,9 @@ resources:
       isReferenced: true
       repeats: true
       additionalSegments:
-             - MSH
-             - PID
-             - PV1
+        - MSH
+        - PID
+        - PV1
              
     - resourceName: Observation
       segment: OBX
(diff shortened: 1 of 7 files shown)


src/main/resources/hl7/message/ADT_A08.yml
```diff
@@ -12,48 +12,60 @@ resources:
...
-             - MSH
+        - MSH
+
+    - resourceName: Procedure
+      segment: .PR1
+      group: PROCEDURE
+      resourcePath: resource/Procedure
+      repeats: true
+      additionalSegments:
+        - .ROL
```
klwhaley, line 68: do we have any support for ROL yet?  maybe thats something done earlier?

src/test/java/io/github/linuxforhealth/hl7/message/HL7ADTMessageTest.java
```diff
@@ -28,403 +30,525 @@
...
+                        .filter(v -> ResourceType.AllergyIntolerance == v.getResource().getResourceType())
+                        .map(BundleEntryComponent::getResource).collect(Collectors.toList());
+        assertThat(allergyIntoleranceResource).hasSize(2); // from AL1
+
+        List<Resource> conditionResource = e.stream()
+                        .filter(v -> ResourceType.Condition == v.getResource().getResourceType())
+                        .map(BundleEntryComponent::getResource).collect(Collectors.toList());
+        assertThat(conditionResource).hasSize(1); // from DG1
+
+        // Confirm that there are no extra resources
```
klwhaley, line 166: should we add a check for document reference and set to 0, but that it should be 1 wiht the content from OBX 'tx' and reference the card?

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -247,102 +216,6 @@ public void test_adt_40_message_with_adt_a39_structure_specified() throws Except
 
   }
 
-  @Test
```
klwhaley, line 250: I thought the point of this test was to make sure a message with no msh 9.3 was parse and converted due to a old bug.  Do we have anohter test for that?

klwhaley (approved):
Approved, with a few questions to consider

src/main/resources/hl7/message/ADT_A08.yml
lisadier: Right, we haven't done any work for ROL, this is the same way it was done in ADT_A01.  It does match the grouping and definition of segments for the message, however it is not currently used in Procedure.yml.  I could go either way on this one.  It isn't wrong yet it isn't yet used in Procedure.  How about we leave it for now, and adjust if needed when we implement PR1 (for Procedures).

src/test/java/io/github/linuxforhealth/hl7/message/HL7ADTMessageTest.java
lisadier: I agree, good idea, done.

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
lisadier: Good catch, I didn't notice that was the point of the test.  I've added more comments to make it more clear why this test is here.

PR #332, opened 2021-11-02
Stop creating unwanted DocRefs

cragun47:

This solves the problem of unwanted DocRefs created when there is no TXA and the OBX's aren't of type TX.

Moves some the logic from the yml to a JEXL util where more complex logic can be applied.

Tests updated to match expected behavior and references to TODO and 849 are removed.


src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
```diff
@@ -104,6 +106,34 @@ public static String getEncounterStatus(Object var1, Object var2, Object var3) {
...
+    // Together with the required:true, this creates logic for messages (MDM, PPR, ORM, OMP) that create DocumentReferences 
+    // a) when there is a TXA, create a DocRef
+    // b) when there is no TXA, and there is an ORC with OBX of type TX, create a DocRef
+    // c) otherwise when there is no TXA create no DocRef
+    // Observation creation is controlled by different code
+    public static String getDocumentReferenceStatus(Object txa, Object txa19, Object orc, Object obr25, Object obx2) {
+        LOGGER.info("Generating DocumentReference status");
+        LOGGER.debug("Generating DocumentReference status from txa{}, txa19 {}, orc {}, obr25 {}, obx2 {}, ", txa, txa19, orc, obr25, obx2);
+
+        if (txa != null || (txa == null && orc!= null && Objects.equals(Hl7DataHandlerUtil.getStringValue(obx2),"TX"))) {
```
LisaWellman, line 121: The check for `txa == null` should not be needed, as if it is not null the first condition will be satisfied.

LisaWellman (approved):

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
cragun47: OK.

PR #335, opened 2021-11-03
add doseQuantity System support and test

jgrant22:
For card [687](https://github.ibm.com/wh-imaging/clinical-data-pipeline/issues/687)
Signed-off-by: jared.grant@ibm.com <jared.grant@ibm.com>


src/test/java/io/github/linuxforhealth/hl7/segments/Hl7ImmunizationFHIRConversionTest.java
```diff
@@ -114,9 +120,14 @@ public void testImmunization() throws IOException {
...
+    assertThat(immunization1.hasDoseQuantity()).isTrue();
+    assertThat(immunization1.getDoseQuantity().getValue().toString()).isEqualTo("0.5");
+    assertThat(immunization1.getDoseQuantity().getUnit()).isEqualTo("ML");
+    assertThat(immunization1.getDoseQuantity().getSystem()).isNull();
   }
   // TODO: 10/15/21 RXA-9 (also mapped to primarySource)
-  //  RXA-6,7 doseQuantity
   //  RXA-18 statusReason
   //  RXA-20 (status, statusReason, isSubpotent)
 }
```
LisaWellman, line 133: These are good test, I think we want a third test that checks a known system -- I would use UCUM, eg
`ml^MilliLiter [SI Volume Units]^UCUM`
We then expect a system of http://unitsofmeasure.org (from CodingSystemMapping.yml)

LisaWellman (changes requested):
Not sure if this will work for all cases, see comment about adding a test to see if it will when the system is known.

LisaWellman (approved):

PR #338, opened 2021-11-10
Fix part of  multiple service requests mixing id values

cragun47:
This fixes service requests mixing values.
It does not fix document references mixing service request values.
A TODO is left in the code for future reference, as well as tests with expected results.

### DIFF
--- a/src/main/resources/hl7/datatype/Identifier_var.yml
+++ b/src/main/resources/hl7/datatype/Identifier_var.yml
@@ -16,10 +16,10 @@ value:
 
 system_1:
    type: STRING
-   valueOf: $systemId 
+   valueOf: $systemId
 
 system_2:
-   condition: $systemCX NOT_NULL
+   condition: $valueIn NOT_NULL && $systemCX NOT_NULL 
    valueOf: $join
    vars:
       join: $systemCX, GeneralUtils.noWhiteSpace(join)
--- a/src/main/resources/hl7/datatype/Identifier_var.yml
+++ b/src/main/resources/hl7/datatype/Identifier_var.yml
@@ -19,7 +19,7 @@ system_1:
    valueOf: $systemId
 
 system_2:
-   condition: $valueIn NOT_NULL && $systemCX NOT_NULL 
+   condition: $systemCX NOT_NULL 
    valueOf: $join
    vars:
       join: $systemCX, GeneralUtils.noWhiteSpace(join)
--- a/src/test/java/io/github/linuxforhealth/hl7/message/Hl7MDMMessageTest.java
+++ b/src/test/java/io/github/linuxforhealth/hl7/message/Hl7MDMMessageTest.java
@@ -7,6 +7,7 @@ package io.github.linuxforhealth.hl7.message;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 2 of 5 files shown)


cragun47:
Tag @LisaWellman for code review.

src/main/resources/hl7/datatype/Identifier_var.yml
```diff
@@ -16,10 +16,10 @@ value:
 
 system_1:
    type: STRING
-   valueOf: $systemId 
+   valueOf: $systemId
 
 system_2:
-   condition: $systemCX NOT_NULL
+   condition: $valueIn NOT_NULL && $systemCX NOT_NULL
```
klwhaley, line 22: not sure I understand why we need valueIn here...is it so we don't end up with a system and no value? but i thought becuase of the checks above this that wouldn't happen, ie we wouldn't get this far if we didn't have valueIn

src/test/java/io/github/linuxforhealth/hl7/message/Hl7MDMMessageTest.java
```diff
@@ -33,6 +39,117 @@
...
+        List<BundleEntryComponent> e = ResourceUtils.createFHIRBundleFromHL7MessageReturnEntryList(hl7message);
+
+        // Check for the expected resources
+        List<Resource> encounterResource = ResourceUtils.getResourceList(e, ResourceType.Encounter);
+        assertThat(encounterResource).hasSize(1); // from EVN, PV1
+
+        List<Resource> patientResource = ResourceUtils.getResourceList(e, ResourceType.Patient);
+        assertThat(patientResource).hasSize(1); // from PID
+
+        // Two ServiceRequests.  One for Dr AAA the other for Dr BBB
```
klwhaley, line 77: comment seems a bit confusing, isn't this because we have two ORCs?

src/test/java/io/github/linuxforhealth/hl7/message/Hl7MDMMessageTest.java
```diff
@@ -33,6 +39,117 @@
...
+            if (ident.hasType()) {
+                String code = ident.getType().getCodingFirstRep().getCode();
+                assertThat(ident.getValue()).hasToString(matchDocRefIdValuesSystems.get(code).get(0));
+                String system = ident.getSystem() == null ? "" : ident.getSystem();
+                assertThat(system).hasToString(matchDocRefIdValuesSystems.get(code).get(1));
+            }
+        }  
+
+        // Confirm that no extra resources are created
+        assertThat(e.size()).isEqualTo(12);
```
klwhaley, line 149: i see us test for 5 of the 12...what are the other 7? can we just do the check its the right resource + number (not go into the resource)

klwhaley (approved):
approve with a few questions

src/main/resources/hl7/datatype/Identifier_var.yml
cragun47: Good catch.  From previous other attempts to solve.  Removed.

src/test/java/io/github/linuxforhealth/hl7/message/Hl7MDMMessageTest.java
cragun47: Clarified:
Two ServiceRequests from the two ORC's.  One references Dr AAA the other references Dr BBB

PR #340, opened 2021-11-12
Eliminate extraneous newline added at end of text reports.

LisaWellman:
The separator string was being added at the end of multi-line reports.  This eliminates that, while preserving the capability to add the separator char after the last element when it is desired.

Also changed remaining tests that output the json to be debug log statement instead of info.

Signed-off-by: Lisa S. Wellman <peace@us.ibm.com>

### DIFF
--- a/src/test/java/io/github/linuxforhealth/hl7/segments/Hl7NoteFHIRConverterTest.java
+++ b/src/test/java/io/github/linuxforhealth/hl7/segments/Hl7NoteFHIRConverterTest.java
@@ -14,9 +14,9 @@ import java.util.Map;
 import java.util.stream.Stream;
 
 import org.hl7.fhir.r4.model.Bundle.BundleEntryComponent;
-import org.hl7.fhir.r4.model.Condition.ConditionEvidenceComponent;
 import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.Condition;
+import org.hl7.fhir.r4.model.Condition.ConditionEvidenceComponent;
 import org.hl7.fhir.r4.model.MedicationRequest;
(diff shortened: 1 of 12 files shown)


src/test/java/io/github/linuxforhealth/hl7/segments/Hl7NoteFHIRConverterTest.java
```diff
@@ -88,8 +88,8 @@ public void testNoteCreationServiceRequestMutipleOBX(String message, int numExpe
         assertThat(serviceRequest.hasNote()).isTrue();
         assertThat(serviceRequest.getNote()).hasSize(1);
         // Processing adds "  \n" two spaces and a line feed
-        assertThat(serviceRequest.getNote().get(0).getText())
-                .isEqualTo("TEST ORC/OBR NOTE AA line 1  \nTEST NOTE AA line 2  \nTEST NOTE AA line 3");
+        assertThat(serviceRequest.getNote().get(0).getTextElement().getValueAsString()).isEqualTo(
```
cragun47, line 91: Do you think we need a comment that we must use  `getTextElement().getValueAsString()` in order to see the String Value we need to test against, rather than the MarkdownType?

Maybe:
`NOTE: the note contains and Annotation, which contains a MarkdownType that has the string.  Must use getTextElement().getValueAsString() to see untrimmed contents.`

cragun47 (approved):
Looks good.  One suggestion for a comment.

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7NoteFHIRConverterTest.java
LisaWellman: Added

PR #341, opened 2021-11-16
 Immunization.statusReason not being converted from RXA.18

jgrant22:
Immunization.statusReason is not being converted from RXA.18 for vxu-v04-sample04.json (when .status="not done"?)
Segment is:
RXA|0|1|20170513|20170513|62^HPV quadrivalent^CVX|999|||||^^^MIICSHORTCODE|||||||00^Parental Refusal^NIP002||RE


src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -504,6 +505,27 @@ else if (value instanceof PPN) {
...
+       // System.out.println("test");
+        if (value instanceof CWE) {
+            CWE cwe = (CWE) value;
+            String actReason = cwe.getCwe1_Identifier().toString();
+            String code = getFHIRCode(actReason, V3ActReason.class);
+            String system = cwe.getCwe3_NameOfCodingSystem().toString();
+            String display = cwe.getCwe2_Text().toString();
+
+            if ( actReason.equals("IMMUNE") || actReason.equals("MEDPREC") || actReason.equals("OSTOCK") ||  actReason.equals("PATOBJ")) {
+                V3ActReason ar = V3ActReason.fromCode(code);
```
jgrant22, line 518: according to the HL7 website there are a few codes we expect from act reason.  If the code is not one of these it goes through coding system v2.

src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
```diff
@@ -106,6 +106,13 @@ V3ActCode:
   P: PRENC
   O: AMB
 
+#Using Act Reason for immunization.statusReason
+V3ActReason:
+  IMMUNE: IMMUNE
+  MEDPREC: MEDPREC
+  OSTOCK: OSTOCK
+  PATOBJ: PATOBJ
```
jgrant22, line 114: I didn't see what mapped on the HL7 side though? in one of the HL7 messages we use has `00^Parental Refusal^NIP002` and expects the code to be `00` and that does not match. the table look up which returns null.

cragun47:
Decision was to NOT do a mapping because the mapping is imperfect and we lose data.  For now we will accept the way the converter creates the CWE as the ground truth.  We will abandon this PR.

Perhaps in the future we will do two reasons, one that preserves the content and one that maps to the best mapping we can for the new codes.
IF we do this, we cannot this PR 341 as is.  
The way to do it in the future is more like:
Do a mapping something like this, or maybe two mappings, one specifically for NIP002
```yaml
V3ActReason:
  IMMUNE: IMMUNE
  MEDPREC: MEDPREC
  OSTOCK: OSTOCK
  PATOBJ: PATOBJ
  00: PATOBJ
  01: RELIG
  03: PATOBJ
```
This code in `SimpleDataValueResolver.java` needs to check for NIP002 first and then map from 00, 01, 03.  All other codings should pass through CWE processing.
```java
    public static final ValueExtractor<Object, SimpleCode> IMMUNIZATION_STATUS_REASON = (Object value) -> {
       // System.out.println("test");
        if (value instanceof CWE) {
            CWE cwe = (CWE) value;
            String actReason = cwe.getCwe1_Identifier().toString();
            String code = getFHIRCode(actReason, V3ActReason.class);
            String system = cwe.getCwe3_NameOfCodingSystem().toString();
            String display = cwe.getCwe2_Text().toString();

            if ( actReason.equals("IMMUNE") || actReason.equals("MEDPREC") || actReason.equals("OSTOCK") ||  actReason.equals("PATOBJ")) {
                V3ActReason ar = V3ActReason.fromCode(code);
                System.out.println(code);
                return new SimpleCode(code,ar.getSystem(),ar.getDisplay());
            }
        }
         return null;
    };
```

This change to `CodeableConcept.yml` should be removed.   We don't want to have a specialized CodeableConcept.  Instead we want the IMMUNIZATION_STATUS_REASON to be processed on the statusReason and give us a mapped value or a pass-through value.
```yaml
coding_6:
   type: IMMUNIZATION_STATUS_REASON
   condition: $customCoding NOT_NULL
   valueOf: $customCoding
   specs: $customCoding
   vars:
      customCoding: CWE

```

In `Immunization.yml` we want something like this which preprocesses RXA.18 to get the right coding.
statusReason_2:
    condition: $coding NOT_NULL
    valueOf: datatype/CodeableConcept
    expressionType: resource
    specs: RXA.18
    vars:
       coding: IMMUNIZATION_STATUS_REASON, RXA.18

PR #344, opened 2021-11-17
add immunization.recorded

jgrant22:
RXA-22 to Immunization.recorded


LisaWellman (approved):
Looks like some bleed through with the statusCode work, ignored that and concentrated on recorded

PR #347, opened 2021-11-22
Fix missing mfg ref  in Imm on multiples

cragun47:

This fixes the missing Immunization manufacturer (and ref Id) when there are multiple RXAs.
Adds tests.

### DIFF
--- a/src/main/resources/hl7/resource/Immunization.yml
+++ b/src/main/resources/hl7/resource/Immunization.yml
@@ -61,9 +61,12 @@ reportOrigin:
    expressionType: resource
    specs: RXA.9
 manufacturer:
+   condition: rxa17 NOT_NULL
    valueOf: resource/Organization
    expressionType: reference
    specs: RXA.17
+   vars: 
+      rxa17: RXA.17
 lotNumber:
    type: STRING
    valueOf: RXA.15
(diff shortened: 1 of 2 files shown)


src/main/resources/hl7/resource/Immunization.yml
```diff
@@ -61,9 +61,12 @@ reportOrigin:
    expressionType: resource
    specs: RXA.9
 manufacturer:
```
LisaWellman, line 63: I don't know enough about the engine or this specific path to know if we could add a check so that this yml condition would not be required.

At the least, can we add something either to the external doc or our internal tips doc about this?  As in when a null check needs to be added -- what is special about this circumstance that makes it needed.

LisaWellman (approved):

src/main/resources/hl7/resource/Immunization.yml
cragun47: It's not completely clear when the specs statement alone was not sufficient.  It may be that some internal buffers are not being correctly reset.  I will mention this to @pbhallam at an upcoming meeting.

cragun47: It's not completely clear why the spec statement alone was not sufficient to handle the null case.  Perhaps some internal problem with buffers being cleared.  I will talk  to @pbhallam at an upcoming meeting and determine if something can be documented.

PR #358, opened 2021-12-08
Fixes for sonarqube code smells + javadoc change to not create index

klwhaley:
1. Change to javadoc not to create index (which in some instances was bringing jszip)
2. Fixes for variety of sonarqube code smells

### DIFF
--- a/src/test/java/io/github/linuxforhealth/hl7/message/SegmentUtilTest.java
+++ b/src/test/java/io/github/linuxforhealth/hl7/message/SegmentUtilTest.java
@@ -6,11 +6,15 @@
 package io.github.linuxforhealth.hl7.message;
 
 import static org.assertj.core.api.Assertions.assertThat;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+
 import org.junit.jupiter.api.Test;
+
 import com.google.common.collect.Lists;
+
 import ca.uhn.hl7v2.HL7Exception;
 import ca.uhn.hl7v2.model.Message;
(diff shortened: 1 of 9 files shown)


src/test/java/io/github/linuxforhealth/hl7/message/SegmentUtilTest.java
```diff
@@ -24,504 +28,446 @@
...
+
+    // Tests for extracting segment from group
+    @Test
+    public void test_single_order_group() throws HL7Exception {
+
+        Message hl7message = getMessage(messageSingleOrderGroup);
+        HL7DataExtractor hl7DTE = new HL7DataExtractor(hl7message);
+        List<SegmentGroup> segmentGroups = SegmentExtractorUtil.extractSegmentGroups(ORDER_GROUP_LIST,
+                "OBX", Lists.newArrayList(), hl7DTE, Lists.newArrayList());
+        assertThat(segmentGroups).isNotEmpty().hasSize(2);
```
LisaWellman, line 106: Wow, surprised it complained about this!

LisaWellman (approved):
Not a big fan of assertThat, to me they just read weird.
Can they be
`assertThat(result.get("use").isNull());`
rather than
`assertThat(result.get("use")).isNull();`
?

Not requiring a change, approving.

src/test/java/io/github/linuxforhealth/hl7/message/SegmentUtilTest.java
klwhaley: if you are asserting on the same field, they need to be chained, not separate lines :)

PR #361, opened 2021-12-09
688 immu reaction detail

jgrant22:


src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
```diff
@@ -108,6 +109,16 @@ public static String getEncounterStatus(Object var1, Object var2, Object var3) {
         return status.toCode();
     }
 
+    public static String handleDate(Object var) {
+        if (var != null){
+            System.out.print(SimpleDataValueResolver.DATE_TIME.apply(var));
+            String date = SimpleDataValueResolver.DATE_TIME.apply(var);
+            System.out.print(Date.of(date));
+            return "2017-10-1";
```
LisaWellman, line 117: clearly this is a testing method, did you mean to deliver?

src/main/resources/hl7/resource/Immunization.yml
```diff
@@ -166,4 +166,17 @@ reasonCode:
...
-   expressionType: HL7Spec
\ No newline at end of file
+   expressionType: HL7Spec
+reaction:
+   valueOf: secondary/Reaction
+   generateList: true
+   expressionType: resource
+   vars:
+      detail: OBX.5
+education:
```
LisaWellman, line 176: This should only be created when OBX.3 has specific values, it looks like the checks in Education are intended to do that, let's add a comment here indicating that.

Will need to verify in unit tests that it is not being created in other cases.  Maybe those are there, but noting that here in case I forget to look for that in the testcases.

src/main/resources/hl7/resource/Immunization.yml
```diff
@@ -166,4 +166,17 @@ reasonCode:
 recorded:
    type: DATE_TIME
    valueOf: RXA.22
-   expressionType: HL7Spec
\ No newline at end of file
+   expressionType: HL7Spec
+reaction:
```
LisaWellman, line 170: This should only be created in certain instances (OBX3 is not one of the 'special' types), it looks like there are no checks to stop it from always being created.
Seems like maybe this should be another specific special case, but I am not seeing any OBX.3 value to check.

src/main/resources/hl7/secondary/Reaction.yml
```diff
@@ -9,3 +9,9 @@ manifestation:
    generateList: true
    expressionType: resource
    specs: AL1.5  
+
+detail:
+   valueOf: resource/Observation
+   expressionType: reference
+   generateList: true
```
LisaWellman, line 16: detail is not a list; reaction is but I think this is making detail a list

src/main/resources/hl7/secondary/Education.yml
```diff
@@ -0,0 +1,42 @@
...
+
+
+reference:
+   condition: $obx3 EQUALS 30956-7
+   valueOf: $reference
+   expressionType: HL7Spec
+   vars:
+      obx5: STRING, OBX.5
+      obx3: STRING, OBX.3.1
+      reference: $id + obx5
```
LisaWellman, line 23: I think for now we assume OBX.5 has the full URL and not append `urn:id:`

src/main/resources/hl7/secondary/Education.yml
```diff
@@ -0,0 +1,42 @@
...
+
+#publicationDate:
+#   condition: $obx3 EQUALS 29768-9
+#   valueOf: "GeneralUtils.handleDate(pubDate)"
+#   expressionType: resource
+#   vars:
+#      pubDate: DATE_TIME, OBX.5
+#      obx3: STRING, OBX.3.1
+
+presentationDate:
```
LisaWellman, line 36: Remove if this is not working

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7ImmunizationFHIRConversionTest.java
```diff
@@ -64,6 +65,7 @@ public void testImmunization() throws IOException {
         assertThat(resource.hasRecorded()).isTrue(); //RXA.22
         assertThat(resource.getRecordedElement().toString().contains("2012-09-01")); //RXA.22
         String manufacturerRef = resource.getManufacturer().getReference();
+        String reactionDetail= resource.getReactionFirstRep().getDetail().getReference();
```
LisaWellman, line 68: This will need to removed once you add checks to reaction.
In this case, the OBX indicates a vaccination info sheet, not a reaction to an immunization.

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7ImmunizationFHIRConversionTest.java
```diff
@@ -107,6 +109,19 @@ public void testImmunization() throws IOException {
         assertThat(org.getId()).isEqualTo(manufacturerRef);
         assertThat(org.hasContact()).isFalse();
 
+        // Looking for one Observation that matches the Reaction.Detail reference
```
LisaWellman, line 112: This also will not be generated on this message once checks are added; there is no indication of a reaction here.  This check will need to be added to a message that has a different code for OBX.3 (not one of the "special" ones)

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7ImmunizationFHIRConversionTest.java
```diff
@@ -120,7 +135,7 @@ public void testImmunization() throws IOException {
                 + "ORC|RE||197027||CP|||||^Clerk^Myron|||||||RI2050\r"
                 + "RXA|0|1|20130531|20130531|48^HIB PRP-T^CVX|0.5|ML^^^||00^new immunization record^NIP001|^Sticker^Nurse|^^^RI2050||||33k2a|20131210|PMC^sanofi^MVX|00^Patient refusal^NIP002|||A\r"
                 + "RXR|C28161^IM^NCIT^IM^INTRAMUSCULAR^HL70162|RT^right thigh^HL70163\r"
-                + "OBX|1|CE|64994-7^vaccine fund pgm elig cat^LN|1|V02^VFC eligible Medicaid/MedicaidManaged Care^HL70064||||||F|||20130531|||VXC40^per imm^CDCPHINVS\r";
+                + "OBX|1|ST|30956-7^vaccine type^LN|1|107||||||F|||20130531|||VXC40^per imm^CDCPHINVS\r";
```
LisaWellman, line 138: Given previous comment, change 107 to be a URL.  I don't know what a valid type would be, so OK leaving as String for now.

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7ImmunizationFHIRConversionTest.java
```diff
@@ -140,14 +155,17 @@ public void testImmunization() throws IOException {
...
+
         // Test should only return RXA.10, ORC.12  is empty
         // If RXA.20 is RE but RXA.18 is blank then we use PATOBJ from v3ActReason
         hl7VUXmessageRep = "MSH|^~\\&|MYEHR2.5|RI88140101|KIDSNET_IFL|RIHEALTH|20130531||VXU^V04^VXU_V04|20130531RI881401010105|P|2.5.1|||NE|AL||||||RI543763\r"
                 + "PID|1||12345^^^^MR||TestPatient^Jane^^^^^L||||||\r"
                 + "ORC|RE||197027||PA|||||^Clerk^Myron|||||||RI2050\r"
                 + "RXA|0|1|20130531|20130531|48^HIB PRP-T^CVX|0.5|ML^^UCUM||00^new immunization record^NIP001|^Sticker^Nurse|^^^RI2050||||33k2a|20131210|PMC^sanofi^MVX||00^refusal|RE|A\r"
                 + "RXR|C28161^IM^NCIT^IM^INTRAMUSCULAR^HL70162|RT^right thigh^HL70163\r"
-                + "OBX|1|CE|64994-7^vaccine fund pgm elig cat^LN|1|V02^VFC eligible Medicaid/MedicaidManaged Care^HL70064||||||F|||20130531|||VXC40^per imm^CDCPHINVS\r";
+                + "OBX|1|DT|29768-9^vaccine fund pgm elig cat^LN|1|20071115011212||||||F|||20130531|||VXC40^per imm^CDCPHINVS\r";
```
LisaWellman, line 168: I would not expect a time on this value, change to just `20071115`

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7ImmunizationFHIRConversionTest.java
```diff
@@ -197,6 +218,10 @@ public void testImmunization() throws IOException {
 
         assertThat(immunization.getStatus().getDisplay()).isEqualTo("completed"); //Status defaults to completed
         assertThat(immunization.hasStatusReason()).isFalse();
+        assertThat(immunization.hasDoseQuantity()).isFalse();
+
+//        //Education.PresentationDate shows under the condition that OBX.3.1 has 29769-7 as the code
```
LisaWellman, line 223: Remove if not working, or add a TODO comment

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
jgrant22: yea I will get rid of this

PR #379, opened 2021-12-30
Issue366 support RelatedPerson from Coverage (IN1)

cragun47:

This adds support for RelatedPerson.  Creates the RelatedPerson if the relationship is not empty or SEL (self).
Creates resource references between the Patient and the Related Person.
Supports special case of relationship SEL.  No RelatedPerson is created, but a reference from subscribed to patient is.

Some reformatting of SimpleDataValueResolver.

### DIFF
--- a/src/main/resources/hl7/resource/RelatedPerson.yml
+++ b/src/main/resources/hl7/resource/RelatedPerson.yml
@@ -1,0 +1,63 @@
+#
+# (C) Copyright IBM Corp. 2020, 2021
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+resourceType: RelatedPerson
+# Represents data that needs to be extracted for a RelatedPerson Resource in FHIR
+# reference: https://build.fhir.org/relatedperson.html
+id:
+   type: STRING
+   valueOf: UUID.randomUUID()
+   expressionType: JEXL
+
+identifier:
+   condition: $valueIn NOT_NULL
+   valueOf: datatype/Identifier_var
+   generateList: true
+   expressionType: resource
+   vars:
+      valueIn: String, IN1.49.1 | $relatedId
+   constants:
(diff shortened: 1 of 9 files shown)


src/main/resources/hl7/resource/RelatedPerson.yml
```diff
@@ -0,0 +1,63 @@
...
+      system: http://terminology.hl7.org/CodeSystem/v2-0203
+      code: XV
+      display: "Health Plan Identifier"
+
+patient: 
+   valueOf: datatype/Reference
+   expressionType: resource
+   generateList: true
+   specs: $patientRef
+
```
jgrant22, line 31: I am just curious, where did the patientRef come from?

jgrant22 (approved):

src/main/resources/hl7/resource/RelatedPerson.yml
cragun47: $patientRef comes from Coverage.yml, which create the RelatedPerson.  Coverage grabs the $Patient and passes it in.   I should see if it comes in for free because it's in scope.

cragun47: @jgrant22  That was a great question.  It turns out that both IN1 and $Patient are in scope, so they do not need to be passed in explicitly.    I removed the explicit passing because it was adding clutter.  I added a comment so it is clearer where they are coming from.   

@LisaWellman You may wish to review this simplification; I think you will agree it's better.

LisaWellman: I like it!  Could help with collisions on var names too.  We might want to document as a best practice on our github page

cragun47: I added a section to best practices.

PR #383, opened 2022-01-05
remove hardcoded unit

jgrant22:
Fix for 705
rateRation had a default unit which was my error. Simple fix.


cragun47 (approved):
Looks correct to me.

Issue #366, opened 2021-12-14
Foundational support of IN1 segment to Coverage resource and DFT_P03 support


cragun47:
Foundational support for IN1 mapped to Coverage

* Create `Coverage` template
   * Map `coverage.identifier`
   * Map `coverage.status`
   * Map `coverage.payor`, create a referenced Organization.
   * Map `coverage.beneficiary`, reference Patient created from PID
   * Map `coverage.class`
   * Map `coverage.order`
   * Map `coverage.subscriberId`
   * Map `coverage.relatedPerson`, create a referenced RelatedPerson
   * Map `coverage.relationship` (to the RelatedPerson)
   * Map reference to Patient when relationship is SEL (self)
* Create `Class` template
* Create `RelatedPerson` template
   * Map `relatedPerson.name`
   * Map `relatedPerson.brithdate`
   * Map `relatedPerson.address`
   * Map `relatedPerson.gender`
   * Map `relatedPerson.relationship`
   * Map `relatedPerson.identifier`
* Map to the referenced Organization:
  * Map `organization.identifier`
  * Map `organization.name`
  * Map `organization.contact` and associated detail: name, period, telecom
  * Map `organization.address` and associated detail including period

Initial support for DFT_P03 message type.  Handle:
  * `MSH`
  * `PID` to Patient
  * `PV1` to Encounter
  * `IN1` to Coverage
  * `ORC` to ServiceRequest
  * `OBX` to Observation
_Note FT1 is not currently utilized or supported._

PR #392, opened 2022-01-11
#366 Improved RelPerson Id

cragun47:

This makes some small but important changes to RelatedPerson identifier so that the display and code are looked up, rather than hard coded.

Adds capability to Identifier_var to use a coding.
Uses CODING_SYSTEM_V2_IDENTIFIER to create a coding.
Added documentation about scope variables for Identifier_var and CodeableConcept_var.

### DIFF
--- a/src/main/resources/hl7/datatype/Identifier_var.yml
+++ b/src/main/resources/hl7/datatype/Identifier_var.yml
@@ -1,12 +1,12 @@
 #
-# (C) Copyright IBM Corp. 2020
+# (C) Copyright IBM Corp. 2020, 2022
 #
 # SPDX-License-Identifier: Apache-2.0
 #
 # Creates an identifier, you must provide all the values and variables, including those for CodeableConcept_var
 ---
 
-# Used when a coding is not availble but separate values are
+# Used when a coding is not available but separate values are
 type_1: 
    condition: $code NOT_NULL && $coding NULL
    valueOf: datatype/CodeableConcept_var
--- a/src/main/resources/hl7/datatype/Identifier_var.yml
+++ b/src/main/resources/hl7/datatype/Identifier_var.yml
@@ -5,10 +5,26 @@
 #
 # Creates an identifier, you must provide all the values and variables, including those for CodeableConcept_var
 ---
-type: 
+
+# Used when a coding is not availble but separate values are
+type_1: 
+   condition: $code NOT_NULL && $coding NULL
    valueOf: datatype/CodeableConcept_var
    generateList: true
    expressionType: resource
+    # CodeableConcept_var uses these scope vars: 
+    # $code (for CodeableConcept.Coding.code)
+    # $system (for CodeableConcept.Coding.system)
+    # $display (for CodeableConcept.Coding.display)
+    # $version (for CodeableConcept.Coding.version)
+    # $text (for CodeableConcept.text)
+
+# Used when a coding is available
+type_2:
+    condition: $coding NOT_NULL && $code NULL 
+    valueOf: datatype/CodeableConcept
+    generateList: true
+    expressionType: resource
     
 value:
    type: STRING
--- a/src/main/resources/hl7/resource/RelatedPerson.yml
+++ b/src/main/resources/hl7/resource/RelatedPerson.yml
@@ -23,11 +23,14 @@ identifier:
    generateList: true
    expressionType: resource
    vars:
+      # For identifier.value
       valueIn: String, IN1.49.1 
-   constants:
-      system: http://terminology.hl7.org/CodeSystem/v2-0203
-      code: XV
-      display: "Health Plan Identifier"
+      # For identifier.system; systemCX will process and remove spaces
+      systemCX: IN1.49.4
+      # For identifier.type, code, system, & display.
+      # Because IN1.49.5 is an ID code, this will create a correct coding
+      # because it knows the table from position and looks up the display
+      coding: CODING_SYSTEM_V2_IDENTIFIER, IN1.49.5
 
 patient: 
    valueOf: datatype/Reference
--- a/src/main/resources/hl7/resource/RelatedPerson.yml
+++ b/src/main/resources/hl7/resource/RelatedPerson.yml
@@ -1,5 +1,5 @@
 #
-# (C) Copyright IBM Corp. 2020, 2021
+# (C) Copyright IBM Corp. 2020, 2022
 #
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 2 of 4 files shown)


src/main/resources/hl7/datatype/Identifier_var.yml
```diff
@@ -5,10 +5,26 @@
 #
 # Creates an identifier, you must provide all the values and variables, including those for CodeableConcept_var
 ---
-type: 
+
+# Used when a coding is not availble but separate values are
```
klwhaley, line 9: nit: available is misspelled

src/main/resources/hl7/resource/RelatedPerson.yml
```diff
@@ -23,11 +23,14 @@ identifier:
...
-   constants:
-      system: http://terminology.hl7.org/CodeSystem/v2-0203
-      code: XV
-      display: "Health Plan Identifier"
+      # For identifier.system; systemCX will process and remove spaces
+      systemCX: IN1.49.4
+      # For identifier.type, code, system, & display.
+      # Because IN1.49.5 is an ID code, this will create a correct coding
+      # because it knows the table from position and looks up the display
+      coding: CODING_SYSTEM_V2_IDENTIFIER, IN1.49.5
```
klwhaley, line 33: so if 49.5 is not filled in,  we won't use 49.1? i'm concerned about that as the examples I have seen have only an IN1.49.1.  We need to cover that scenario too...

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7FinancialInsuranceTest.java
```diff
@@ -287,8 +287,12 @@ void testInsuranceCoverageByRelatedFields(String messageType) throws IOException
                 // IN1.43 to RelatedPerson.gender
                 // IN1.46 to Identifier 3
                 // IN1.49 to RelatedPerson.identifier
+                //    IN1.49.1 to RelatedPerson.identifier.value
+                //    IN1.49.4 to RelatedPerson.identifier.system
+                //    IN1.49.5 to RelatedPerson.identifier.type.code (must be from terminology.hl7.org/3.0.0/CodeSystem-v2-0203.html)
+                //      NOTE: Purposely XX to ensure we are doing a lookup, and not getting bleed from other hard-coded XV uses
                 // IN1.50 through IN1.53 NOT REFERENCED
-                + "|MEMBER36|||||||F|||Value46|||J494949||||\n";
+                + "|MEMBER36|||||||F|||Value46|||J494949^^^Large HMO^XX||||\n";
```
klwhaley, line 295: we need a test with only 49.1

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7FinancialInsuranceTest.java
```diff
@@ -322,10 +326,10 @@ void testInsuranceCoverageByRelatedFields(String messageType) throws IOException
...
         // Check RelatedPerson identifier
         assertThat(related.getIdentifier()).hasSize(1);
-        assertThat(related.getIdentifier().get(0).getValue()).isEqualTo("J494949"); // IN1.49
-        assertThat(related.getIdentifier().get(0).getSystem()).isNull();
-        DatatypeUtils.checkCommonCodeableConceptAssertions(related.getIdentifier().get(0).getType(), "XV",
-                "Health Plan Identifier",
+        assertThat(related.getIdentifier().get(0).getValue()).isEqualTo("J494949"); // IN1.49.1
+        assertThat(related.getIdentifier().get(0).getSystem()).isEqualTo("urn:id:Large_HMO"); // IN1.49.4
+        DatatypeUtils.checkCommonCodeableConceptAssertions(related.getIdentifier().get(0).getType(), "XX", // IN1.49.5
+                "Organization identifier",
```
klwhaley, line 332: i need help with where 'Organization identifier' is from?

klwhaley (changes requested):

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7FinancialInsuranceTest.java
cragun47: 'Organization identifier' is the real display value for real code 'XX'.  It was looked up from the code in the conversion.  That's why we're using this test.

klwhaley: ahh! i see sorry missed that it was up in the hl7 message...thx

src/main/resources/hl7/datatype/Identifier_var.yml
cragun47: fixed

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7FinancialInsuranceTest.java
cragun47: Comment added to clarify.

src/main/resources/hl7/resource/RelatedPerson.yml
cragun47: We always use IN1.49.1 if it is present as the identifier.value.  If there is no IN1.49.5, we don't create identifier.type.  Only identifier.value is required.

I checked and for this in IN1.49 `|J494947|`...

We get this result:
```json
      "identifier": [ {
        "value": "J494947"
      } ],
```

So the scenario is covered.  Do you feel you need a test for it?  Might be a little overkill.  @klwhaley

klwhaley: could we add a comment to the end of line 33 # coding for identifier.type

klwhaley (approved):
Approving per discussion that 49.1 will create a identifier, 49.5 is only for the type

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7FinancialInsuranceTest.java
cragun47: Discussed offline with Kathryn.  `Value` is required for `identifier`.  When IN1.49.1 is present, `identifier.value` will alway be created, regardless of whether IN1.49.4 or IN1.49.5 are present and create the `identifier.system` and `identifier.type` respectively.

We decided an additional test was not needed, but documentation was added to the .yml.

src/main/resources/hl7/resource/RelatedPerson.yml
cragun47: Decided documentation was sufficient without an additional test.

PR #394, opened 2022-01-12
Return FHIR Bundle object without encoding as string

ayushgarg0694:
- Added a new function to API that returns the FHIR Bundle `org.hl7.fhir.r4.model.Bundle` as is (does not encode to String)
- Refactored validation into its own method
- other minor refactorings

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ The converter supports the following message types/events:
 * ADT_A08 - Patient Administration: Update Patient Information
 * ADT_A34 - Patient Administration: Merge Patient Information - Patient ID Only
 * ADT_A40 - Patient Administration: Merge Patient - Patient Identifier List
+* DFT_P03 - Post Detail Financial Transaction (does not convert FT1)
 * MDM_T02 - Original Document Notifcication and Content
 * MDM_T06 - Document Addendum Notification and Content
 * OMP_O09 - Pharmacy/Treatment Order
@@ -23,6 +24,7 @@ The converter supports the following message segments:
 * AL1 - Patient Allergy Information
 * DG1 - Diagnosis
 * EVN - Event Type
+* IN1 - Insurance
 * MRG - Merge Patient Information
 * MSH - Message Header
 * NTE - Notes and Comments
--- a/src/main/resources/config.properties
+++ b/src/main/resources/config.properties
@@ -1,4 +1,4 @@
 base.path.resource=
-supported.hl7.messages=ADT_A01, ADT_A08, ADT_A34, ADT_A40, MDM_T02, MDM_T06, ORM_O01, OMP_O09, ORU_R01, PPR_PC1, RDE_O11, RDE_O25, VXU_V04
+supported.hl7.messages=ADT_A01, ADT_A08, ADT_A34, ADT_A40, DFT_P03, MDM_T02, MDM_T06, ORM_O01, OMP_O09, ORU_R01, PPR_PC1, RDE_O11, RDE_O25, VXU_V04
 default.zoneid=+08:00
 additional.conceptmap.file=
--- a/src/main/resources/hl7/message/DFT_P03.yml
+++ b/src/main/resources/hl7/message/DFT_P03.yml
@@ -1,0 +1,72 @@
+#
+# (C) Copyright IBM Corp. 2022
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# FHIR Resources to extract from DFT_P03 message
+# 
+################################
+# NOTE: FT1 is not currently utilized. FT1 records are ignored.
+---
+resources:
+  - resourceName: MessageHeader
+    segment: MSH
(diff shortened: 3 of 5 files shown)


src/main/java/io/github/linuxforhealth/hl7/HL7ToFHIRConverter.java
```diff
@@ -136,6 +136,36 @@ public String convert(String hl7MessageData, ConverterOptions options) {
...
+        Preconditions.checkArgument(options != null, "options cannot be null.");
+        FHIRContext context = new FHIRContext(options.isPrettyPrint(), options.isValidateResource());
+        HL7MessageEngine engine = new HL7MessageEngine(context, options.getBundleType());
+
+        Message hl7message = getHl7Message(hl7MessageData);
+        if (hl7message != null) {
+            String messageType = HL7DataExtractor.getMessageType(hl7message);
+            HL7MessageModel hl7MessageTemplateModel = messagetemplates.get(messageType);
+            if (hl7MessageTemplateModel != null) {
+                return hl7MessageTemplateModel.convertToBundle(hl7message, engine);
```
LisaWellman, line 160: Except for this line, this is an exact copy of `convert(String hl7MessageData, ConverterOptions options)` ?  I would expect that, but please refactor so we are not duplicating code

src/main/java/io/github/linuxforhealth/hl7/message/HL7MessageModel.java
```diff
@@ -65,6 +75,48 @@ public String convert(String message, MessageEngine engine) throws IOException {
...
+    } catch (HL7Exception e) {
+      throw new IllegalArgumentException("Cannot parse the message.", e);
+    } finally {
+      if (hparser != null) {
+        hparser.getContext().close();
+      }
+    }
+  }
+
+  public Bundle convertToBundle(Message message, MessageEngine engine){
```
LisaWellman, line 98: I expected that we would need to change `convert(Message message, MessageEngine engine)` to just not do the encodeResourceToString() call.  I would expect to see that method changed to call this method, and then just do the encodeResourceToString on the result before returning so we do not have duplicate code.

src/main/java/io/github/linuxforhealth/hl7/message/HL7MessageModel.java
```diff
@@ -65,6 +75,48 @@ public String convert(String message, MessageEngine engine) throws IOException {
 
   }
 
+  public Bundle convertToBundle(String message, MessageEngine engine) throws IOException{
+    Preconditions.checkArgument(StringUtils.isNotBlank(message),
+            "Input Hl7 message cannot be blank");
+    HL7HapiParser hparser = null;
+    try {
```
LisaWellman, line 82: change convert(String message, MessageEngine engine) to call this method and then convert to a String to avoid duplicate code.

src/main/java/io/github/linuxforhealth/hl7/message/HL7MessageModel.java
```diff
@@ -65,6 +75,48 @@ public String convert(String message, MessageEngine engine) throws IOException {
...
+      hparser = new HL7HapiParser();
+
+      Message hl7message = hparser.getParser().parse(message);
+      return convertToBundle(hl7message, engine);
+
```
LisaWellman, line 87: nit: please remove extraneous newlines

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -15,20 +15,13 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
+import ca.uhn.fhir.context.FhirContext;
 import org.apache.commons.io.IOUtils;
 import org.hl7.fhir.exceptions.FHIRException;
 import org.hl7.fhir.instance.model.api.IBaseResource;
-import org.hl7.fhir.r4.model.Bundle;
+import org.hl7.fhir.r4.model.*;
```
LisaWellman, line 22: Please revert to specific imports.

src/test/java/io/github/linuxforhealth/FHIRConverterTest.java
```diff
@@ -50,6 +43,26 @@
     private static final ConverterOptions OPTIONS = new Builder().withValidateResource().withPrettyPrint().build();
     private static final Logger LOGGER = LoggerFactory.getLogger(FHIRConverterTest.class);
 
+    @Test
+    void test_patient_encounter_bundle_Return() throws IOException {
+
+        String hl7message = "MSH|^~\\&|SE050|050|PACS|050|20120912011230||ADT^A01|102|T|2.6|||AL|NE|764|ASCII||||||^4086::132:2A57:3C28^IPv6\r"
```
LisaWellman, line 49: Simplify the message segments since we are not doing a deep verification.  See other testcases for examples.

PR #396, opened 2022-01-12
Bundle return (#4)

ayushgarg0694:
* refactoring and adding new function to return Bundle

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/HL7ToFHIRConverter.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/HL7ToFHIRConverter.java
@@ -127,6 +127,37 @@ public class HL7ToFHIRConverter {
             String messageType = HL7DataExtractor.getMessageType(hl7message);
             HL7MessageModel hl7MessageTemplateModel = messagetemplates.get(messageType);
             if (hl7MessageTemplateModel != null) {
+                Bundle bundle = hl7MessageTemplateModel.convert(hl7message, engine);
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/linuxforhealth/hl7/HL7ToFHIRConverter.java
```diff
@@ -116,11 +116,25 @@ public String convert(String hl7MessageData) {
...
+     * @return Bundle {@link Bundle} resource.
+     * @throws UnsupportedOperationException - if message type is not supported
+     */
+    public Bundle convertToBundle(String hl7MessageData, ConverterOptions options) {
         Preconditions.checkArgument(StringUtils.isNotBlank(hl7MessageData),
                 "Input HL7 message cannot be blank");
-        Preconditions.checkArgument(options != null, "options cannot be null.");
-        FHIRContext context = new FHIRContext(options.isPrettyPrint(), options.isValidateResource());
-        HL7MessageEngine engine = new HL7MessageEngine(context, options.getBundleType());
+        HL7MessageEngine engine = getMessageEngine(options);
```
LisaWellman, line 137: I would like to get rid of this duplicate call when going down the "string" path.
Can we pass a third parm `HL7MessageEngine engine` to this method and then only do this if null:
`if (engine==null) { engine = getMessageEngine(options); }`

LisaWellman (commented):
Looks good, just one request to avoid duplicate processing in the "string" path

LisaWellman (approved):

PR #402, opened 2022-01-14
add EVN.2 to Encounter.period & add unit test for the different use cases

jgrant22:


src/test/java/io/github/linuxforhealth/hl7/segments/Hl7EncounterFHIRConversionTest.java
```diff
@@ -20,20 +20,9 @@
 import java.util.stream.Collectors;
 
 import org.hl7.fhir.instance.model.api.IBaseResource;
-import org.hl7.fhir.r4.model.Bundle;
+import org.hl7.fhir.r4.model.*;
```
cragun47, line 23: Use explicit imports.  Don't use *.

cragun47 (approved):
Please use explicit class imports.  Otherwise fine.

src/test/java/io/github/linuxforhealth/hl7/segments/Hl7EncounterFHIRConversionTest.java
jgrant22: Is there a way to stop this? it automatically does it and I never catch it

PR #405, opened 2022-01-18
Correct relationship & other in1 tweaks

cragun47:
This   Maps 0063 codes in IN1.17 to values in http://terminology.hl7.org/CodeSystem/v3-RoleCode for RelatedPerson.relationship.

 Maps 0063 to values in http://terminology.hl7.org/CodeSystem/subscriber-relationship for Coverage.relationship
 
 Fixes some comments and other typo errors / omissions.


src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -356,20 +357,31 @@
...
             V3RoleCode relationship = V3RoleCode.fromCode(code);
             return new SimpleCode(code, relationship.getSystem(), relationship.getDisplay(), version);
         } else {
-            // If code is null, it means the code wasn't known in our table, and can't be looked up.
-            // Make a message in the display.
-            String theSystem = V3RoleCode.PRN.getSystem();
-            return new SimpleCode(null, theSystem,
-                    String.format(INVALID_CODE_MESSAGE_FULL, val, theSystem, text, version));
+            // If code is not found in our mapping, return the code itself with no system or display. 
+            return new SimpleCode(val, null, null, null);
```
LisaWellman, line 370: Is this what we think is most consistent with what we have today?  I was looking in SimpleDataValueResolver and it seems like there is a variety of behaviors when the lookup fails, so I am wondering if we retain the system and display from the HL7 message?
Same question on line 384.

LisaWellman (approved):
One question, but looks good in general.  Approving

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
cragun47: I guess we need to talk again.  My understanding is we didn't want to retain the system without a lookup to ensure the system was validated.  And we didn't want to add that complexity in this card.  @LisaWellman

PR #406, opened 2022-01-18
fix data bleed found by Anna and apply changes

jgrant22:
While working on the ground truth for [344](https://github.ibm.com/WH-Imaging/Clinical-Data-Pipeline/issues/344), Anna discovered a scenario where an Identifier with no system still had a system. 

I don't full understand how that was happening but I was able to get a fix for it by moving the nowhitespace function up a level. This did require alot of changes to the yamls. 

There are test in IdentifierFhirConverterTest that cover the case that was failing so I just made the changes and made sure they still worked correctly

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
@@ -21,8 +21,12 @@ import java.time.temporal.UnsupportedTemporalTypeException;
 import ca.uhn.hl7v2.model.v26.datatype.CWE;
 import ca.uhn.hl7v2.model.v26.datatype.PPN;
 import ca.uhn.hl7v2.model.v26.datatype.XCN;
-import ca.uhn.hl7v2.model.v26.segment.PV1;
+import ca.uhn.hl7v2.model.v26.datatype.CX;
+import ca.uhn.hl7v2.model.v26.datatype.EI;
 import ca.uhn.hl7v2.model.v26.datatype.DTM;
+import ca.uhn.hl7v2.model.v26.segment.PV1;
+import ca.uhn.hl7v2.model.v26.segment.IN1;
+
 
 import ca.uhn.hl7v2.model.Varies;
 import org.apache.commons.lang3.BooleanUtils;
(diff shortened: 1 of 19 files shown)


src/test/java/io/github/linuxforhealth/hl7/message/Hl7MDMMessageTest.java
```diff
@@ -123,26 +123,29 @@ void noMixingOfParamsInDocumentReference(String message) throws IOException {
...
+//            Identifier ident = identifiers.get(idIndex);
+//            System.out.println(ident.getType().getCoding());
+//            if (ident.hasType()) {
+//                String code = ident.getType().getCodingFirstRep().getCode();
+//                assertThat(ident.getValue()).hasToString(matchDocRefIdValuesSystems.get(code).get(0));
+//                String system = ident.getSystem() == null ? "" : ident.getSystem();
+//                System.out.println(system + 1);
+//                assertThat(system).hasToString(matchDocRefIdValuesSystems.get(code).get(1));
+//            }
+//        }
```
jgrant22, line 148: I didn't understand this commented out portion. but It had a todo about data bleeding so I am wondering if my fix is the reason this block of code was failing and not returning a system

src/main/java/io/github/linuxforhealth/hl7/data/SimpleDataValueResolver.java
```diff
@@ -390,6 +394,35 @@
...
+        }
+        if (val != null) {
+            String newVal = val.replaceAll("\\s", "_");
+            return "urn:id:" + newVal;
+        }
+
+        return null;
+
+    };
+
```
cragun47, line 425: I talked with Jared about this by Webex.  My concern is that we're getting a field level object and having to per object type do special handling for the individual types.

In the yaml we used to have processing at a subfield: `orgIdSystem: String, IN1.3.4`, now we have processing at a field level `systemCX: NO_WHITESPACE, IN1.3`.

I am not concerned about moving the processing to remove whitespace from a GeneralUtil to a DataValueResolver.  I am concerned about operations for subfields by passing fields.

Jared agreed to see if we can pass a subfield to the DataValueResolver.   I'm not going to do a deeper review until we see if that can be accomplished.

@LisaWellman

cragun47 (changes requested):
As discussed by Webex, please see if we can change DataValueResolver to the subfield level, instead of the field level.

PR #407, opened 2022-01-19
…

jgrant22:
Data bleed caused by missing segment in the message yaml. 
Same thing for Encounter.diagnosis.use

This PR includes fixes for the two threads in slack from the test team.
Both fixes were simple I completely over thought the first issue which is why I moved it here to start clean.

**Problem 1**: One of our testers discovered a scenario where an Identifier with no system still had a system.  If PV1.4 is available then we use it for the system in this case, PV1.4 was not available but we still got a system
```
what we got:
value: 1001918
system: urn:id:1001918
expected:
value: 1001918
```

**Problem 2**:  One of our testers discovered a scenario we were missing a few mappings and that Encounter.diagnosis.use was not cycling through each DG1 segment

### DIFF
--- a/src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
+++ b/src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
@@ -350,10 +350,8 @@ V3ReligiousAffiliation:
   # Unknown
   # VAR:   NO REASONABLE MATCH
 
-DiagnosisRole:
+DiagnosisRole: # maps DG1.6 to Encounter.diagnosis.use
    A: AD
-   F: DD # F stands for final DD stands for discharge Diagnosis which is the final diagnosis before being discharged
-   W: CC # When the patient presents with a chief complaint, such as cough, the provider considers the most common diseases that present with cough, forming a working differential diagnosis list
 
 MedicationRequestStatus:
    AF: active
--- a/src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
+++ b/src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
@@ -352,6 +352,8 @@ V3ReligiousAffiliation:
 
 DiagnosisRole:
    A: AD
+   F: DD # F stands for final DD stands for discharge Diagnosis which is the final diagnosis before being discharged
+   W: CC # When the patient presents with a chief complaint, such as cough, the provider considers the most common diseases that present with cough, forming a working differential diagnosis list
 
 MedicationRequestStatus:
    AF: active
--- a/src/main/resources/hl7/message/ADT_A03.yml
+++ b/src/main/resources/hl7/message/ADT_A03.yml
@@ -30,18 +30,20 @@ resources:
       repeats: false
       isReferenced: true
       additionalSegments:
-             - PV2
-             - EVN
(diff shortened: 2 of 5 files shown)


src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
```diff
@@ -352,6 +352,8 @@ V3ReligiousAffiliation:
 
 DiagnosisRole:
    A: AD
+   F: DD # F stands for final DD stands for discharge Diagnosis which is the final diagnosis before being discharged
+   W: CC # When the patient presents with a chief complaint, such as cough, the provider considers the most common diseases that present with cough, forming a working differential diagnosis list
```
jgrant22, line 357: I left comments here to explain my logic. definitely open for discussion or even removal

src/main/resources/hl7/message/ADT_A03.yml
```diff
@@ -30,18 +30,20 @@ resources:
...
       isReferenced: true
       additionalSegments:
-             - PV2
-             - EVN
-             - MSH
-             - OBX
+        - PV2
+        - EVN
+        - MSH
+        - DG1
```
cragun47, line 36: ADT_A04, ADT_A28, ADT_A31 look like they might fail the same way?
@LisaWellman

cragun47 (approved):
1). Please adjust PR description so that external visitors have access to all links and desicriptions.

2).  Pleas check ADT_A04, ADT_A28, ADT_A31.  They look like the _might_ have the same issue from cut and paste.   Correct as needed, update tests.

src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
```diff
@@ -352,6 +352,8 @@ V3ReligiousAffiliation:
 
 DiagnosisRole:
```
LisaWellman, line 353: Can you add a comment this maps from DG1.6 to Encounter.diagnosis.use

src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
LisaWellman: I would not map F, W, but leave them intact in the JSON (the JSON should contain the original code).  This is a general design change we have been discussing, it should likely be deferred until we tackle it holistically.  If we do defer, this issue will be an outstanding issue (that 'F' is dropped).

src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
jgrant22: added!

src/main/resources/hl7/message/ADT_A03.yml
jgrant22: updated the messages!

src/main/resources/hl7/codesystem/v2ToFhirMapping.yml
LisaWellman: This is what is in the mapping document:
```
F,W have to FHIR equivalent, use the source system http://terminology.hl7.org/CodeSystem/v2-0052
and display (Final, Working)
```

PR #408, opened 2022-01-20
change Encounter.diagnosis.use to use table lookup instead of mapping

jgrant22:
Quick fix, use table look up instead of mappings for Encounter.diagnosis.use


cragun47 (approved):
Looks good to me!

PR #419, opened 2022-01-25
added support for referencing resource

pbhallam:
Signed-off-by: Pushpa Bhallamudi <pbhallam@us.ibm.com>

### DIFF
--- a/TECHNIQUES.md
+++ b/TECHNIQUES.md
@@ -184,5 +184,37 @@ encounter:
   expressionType: resource
   specs: $encounterRef
 ```
+#### Referencing resource values that are created multiple times
+In [Encounter.diagnosis](./src/main/resources/hl7/resource/Encounter.yml) we reference condition which can be created multiple times. In order to account for this we added a function in general utils that can extract each condition that is created, creates an id, and compares that ID in the yaml to make sure the conditions have the proper IDs
 
-
+```yaml
+diagnosis:
+   expressionType: nested
+   evaluateLater: true
+   generateList: true
+   specs: DG1
+   vars:
+      refDG13: BUILD_IDENTIFIER_FROM_CWE, DG1.3
+   expressionsMap:
+     #SPECS: Gets entire list of condition resources
+     #REFCONDITIONID: returns a reference ID, in this case we are return for the value of the identifier that uses "urn:id:extID"
+     #BASEVALUE: in this case base Value will be a condition resource from the specs list of conditions
+     #REFDG13: The identifier value we compare against to make sure the conditions are in the proper order
+     condition:
+        valueOf: datatype/Reference
+        expressionType: resource
+        condition: $refconditionId EQUALS_STRING $refDG13
+        specs: $Condition
+        vars:
+          refconditionId: $BASE_VALUE, GeneralUtils.extractAttribute(refconditionId,"$.identifier[?(@.system==\"urn:id:extID\")].value","String")
+     use:
+        valueOf: datatype/CodeableConcept
+        expressionType: resource
+        specs: DG1.6
+     rank:
+        type: STRING
+        valueOf: DG1.15
+        expressionType: HL7Spec
+```
+`$.identifier[?(@.system==\"urn:id:extID\")].value` allows us to parse through the resource in json format, and we can search for specific values
+in this case we wanted to look at the list of identifiers and get the value of the one that has `urn:id:extID` as the System.
--- a/TECHNIQUES.md
+++ b/TECHNIQUES.md
@@ -185,7 +185,17 @@ encounter:
   specs: $encounterRef
 ```
 #### Referencing resource values that are created from repeating segments
-In [Encounter.diagnosis](./src/main/resources/hl7/resource/Encounter.yml) we reference condition which can be created multiple times by repeating DG1 segments. In order to account for this scenario and possibly others like it, we added a function in general utils that extracts each resource value from the list of resources created from a specific segment(in our case its DG1), creates an id, and compares that ID in the yaml to make sure we are dealing with the specific resource value.
+In cases where we need to match specific resources that are in a list from a `spec`, we can use a nested structure where the outer part identifies the element to match, and the inner part uses condition(s) to find the matching element.
+A good example of this is Encounter.diagnosis in Encounter.yml.
+
+- Outer Structure  
+  - Nested expression type.
+  - Cycles through each DG1 segment via `specs: DG1`.
+  - Creates an identifier in `$refDG13` from DG1.3 for each time we cycle through the `specs: DG1`.
+- Inner Structure (`expressionsMap`)
+  - Cycles through each `specs: $Condition` resource created from each DG1 segment processed in the parent message `ADT_A03`.
+  - Matches the `$refDG13` Identifier from the outer structure to the `$refconditionId` condition identifier from the inner structure.
+  - `$refconditionId` is found in the `$Condition` by using `GeneralUtils.extractAttribute`, which uses a pattern matching utility to find the identifier.
 
 ```yaml
 diagnosis:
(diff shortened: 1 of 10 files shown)


src/main/resources/hl7/resource/Encounter.yml
```diff
@@ -129,11 +129,20 @@ diagnosis:
    evaluateLater: true
```
LisaWellman, line 129: Let's add a comment before line 127 that describes what is happening at a high level.  Something like this, though I don't think I am explaining the "lists" well, I hope you can improve upon it!

# Diagnosis is a list of References to each Condition.
# To get the correct Condition from the Condition list in the expressionsMap spec, we use the Condition.identifier generated from DG1-3.

src/main/resources/hl7/resource/Encounter.yml
```diff
@@ -129,11 +129,20 @@ diagnosis:
...
    generateList: true
    specs: DG1
+   vars:
+      refDG13: BUILD_IDENTIFIER_FROM_CWE, DG1.3
    expressionsMap:
```
LisaWellman, line 134: correct these names -- they are not all uppercase and BASE_VALUE has an underscore

I prefer to have a space after the pound sign before the comment text starts.

src/main/resources/hl7/resource/Encounter.yml
```diff
@@ -129,11 +129,20 @@ diagnosis:
...
+     #SPECS: Gets entire list of condition resources
+     #REFCONDITIONID: returns a reference ID, in this case we are return for the value of the identifier that uses "urn:id:extID"
+     #BASEVALUE: in this case base Value will be a condition resource from the specs list of conditions
+     #REFDG13: The identifier value we compare against to make sure the conditions are in the proper order
```
LisaWellman, line 138: I would not say this is `to make sure the conditions are in the proper order`.  It is to match to the Condition being used for the reference (inner loop) matches the Condition in the outer loop

TECHNIQUES.md
```diff
@@ -184,5 +184,37 @@ encounter:
   expressionType: resource
   specs: $encounterRef
 ```
+#### Referencing resource values that are created multiple times
```
LisaWellman, line 187: I think we need an introductory line here explaining what the scenario is and the solution

src/main/resources/hl7/resource/Encounter.yml
jgrant22: fixed

src/main/resources/hl7/resource/Encounter.yml
jgrant22: fixed

src/main/resources/hl7/resource/Encounter.yml
jgrant22: added

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
```diff
@@ -64,34 +64,30 @@ private Hl7RelatedGeneralUtils() {
...
+        try {
+            String json = ObjectMapperUtil.getJSONInstance().writeValueAsString(data);
 
+            List<?> val = JsonPath.parse(json).read(path, List.class);
+            ValueExtractor<Object, ?> resolver = SimpleDataTypeMapper.getValueResolver(klass);
+            LOGGER.debug("extracted value after json evaluation: {}", val);
+            return resolver.apply(val);
 
+        } catch (RuntimeException | JsonProcessingException e) {
+            LOGGER.warn(
```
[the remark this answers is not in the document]
jgrant22: fixed, thanks

TECHNIQUES.md
```diff
@@ -184,5 +184,37 @@ encounter:
...
+In [Encounter.diagnosis](./src/main/resources/hl7/resource/Encounter.yml) we reference condition which can be created multiple times. In order to account for this we added a function in general utils that can extract each condition that is created, creates an id, and compares that ID in the yaml to make sure the conditions have the proper IDs
 
-
+```yaml
```
LisaWellman, line 190: I am wary of copying the yml here as it will be out of date when it changes.
I think we should reference this JSON as an example, but not copy it here, or at the most copy relevant pieces critical to the explanation.

TECHNIQUES.md
```diff
@@ -184,5 +184,37 @@ encounter:
...
+diagnosis:
+   expressionType: nested
+   evaluateLater: true
+   generateList: true
+   specs: DG1
+   vars:
+      refDG13: BUILD_IDENTIFIER_FROM_CWE, DG1.3
+   expressionsMap:
+     #SPECS: Gets entire list of condition resources
+     #REFCONDITIONID: returns a reference ID, in this case we are return for the value of the identifier that uses "urn:id:extID"
```
cragun47, line 200: nit:  fix "we are return" grammar

TECHNIQUES.md
```diff
@@ -184,5 +184,37 @@ encounter:
...
+     #BASEVALUE: in this case base Value will be a condition resource from the specs list of conditions
```
cragun47, line 201: nit: refer to BASE_VALUE

TECHNIQUES.md
```diff
@@ -184,5 +184,37 @@ encounter:
...
+     use:
+        valueOf: datatype/CodeableConcept
+        expressionType: resource
+        specs: DG1.6
+     rank:
+        type: STRING
+        valueOf: DG1.15
+        expressionType: HL7Spec
+```
+`$.identifier[?(@.system==\"urn:id:extID\")].value` allows us to parse through the resource in json format, and we can search for specific values
```
LisaWellman, line 219: Let's move 219-220 to Encounter.yml.  This is important detail for that specific piece.

TECHNIQUES.md
```diff
@@ -184,5 +184,37 @@ encounter:
...
+   evaluateLater: true
+   generateList: true
+   specs: DG1
+   vars:
+      refDG13: BUILD_IDENTIFIER_FROM_CWE, DG1.3
+   expressionsMap:
+     #SPECS: Gets entire list of condition resources
+     #REFCONDITIONID: returns a reference ID, in this case we are return for the value of the identifier that uses "urn:id:extID"
+     #BASEVALUE: in this case base Value will be a condition resource from the specs list of conditions
+     #REFDG13: The identifier value we compare against to make sure the conditions are in the proper order
```
cragun47, line 202: "The identifier value we compare against to make sure we are dealing with the specific Diagnosis."  I don't think order has anything to do with it.

TECHNIQUES.md
cragun47, line 219: `GeneralUtils.extractAttrtibute` uses utility json-path to parse through the resource in json format, and we can search for specific values.  ` $.identifier[?(@.system==\"urn:id:extID\")].value`  is the json-path notation to find from root ($) identifiers that have a system with "urn:id:extID"

TECHNIQUES.md
jgrant22: fixed

TECHNIQUES.md
jgrant22: fixed

TECHNIQUES.md
jgrant22: fixed

TECHNIQUES.md
jgrant22: fixed

jgrant22: moved

TECHNIQUES.md
jgrant22: done!

TECHNIQUES.md
jgrant22: fixed!

cragun47 (approved):
Much better, thanks.

PR #431, opened 2022-02-02
Lower case Oganization ids

cragun47:

Small change that makes Organization ID's lower case.   Plus tests.


LisaWellman (approved):
Approving, as this solves the casing.
We do need to also handle invalid characters are well, replacing them with an underscore.

Issue #414, opened 2022-01-21
Multiple OBX records need to group by OBX.4, but don't


cragun47:
( This bug was found in "add mappings to Immunization from OBX" #400  That item was closed and this bug remains in the work.  )

The expectation is the multiple OBX’s are combined to one component.  See NHIIS Implemenation guide page 21: https://www.dhhs.nh.gov/dphs/immunization/documents/nhiis-delta-guide-hl7.pdf

We need a way to `spec: OBX` and also inner group on `OBX.4`

The OBX’s are “connected” by the reference string in OBX.4 
![image (14)](https://user-images.githubusercontent.com/4914849/150541554-2f333b60-8c7a-41a4-a3ce-1c21fe88fa9b.png)
Here is another example.  You can see looking at OBX.4 that there are three _sets_ of records.
![image (15)](https://user-images.githubusercontent.com/4914849/150541846-38549859-f17e-4df7-be7e-b7cdf8a33393.png)
This is consistent with CDC documents: https://www.cdc.gov/vaccines/programs/iis/technical-guidance/downloads/hl7-clarification-R6.pdf. Pages 5 & 6
VIS Conformance Statement
 OBX conformance statements
 Clarification
 IZ-24: If RXA-9.1 is valued “00” and RXA-5.1 is valued with a CVX code from table PHVS_VISVaccines_IIS (See Appendix A) then there SHALL BE
an OBX segment with OBX-3.1 valued
“64764-9" (bar
 coded) and one OBX with OBX-3.1 valued “29769-7” (presentation /delivery date) associated. Both OBX shall have the same value in OBX-4
OR
an OBX segment with OBX-3.1
valued “30956-7" (vaccine type) and an OBX segment with OBX-3.1 valued “29768-9” (version date) and one OBX with OBX-3.1 valued “29769-7" (presentation /delivery date) associated. Both OBX shall have the same value in OBX-4

cragun47:
This issue was opened to find a general solution to situations similar to `Immunization.education` multiple records.  However no other needs were found for a general case.   Instead a specific solution was found and captured in  #449.  It can be used as a pattern for future similar needs.

No further action is planned on this issue as described.  Closing.

PR #434, opened 2022-02-03
Issue414 grouping related obx records to one element

cragun47:
@pbhallam  This has tests that show what we need to do.   Immunization.yml also contains two failing efforts to solve the problem.


cragun47:
See explanations in #414 and #449.  This general purpose solution is no longer needed.  A specific solution is now available (#449) The specific solution goes a different direction than this branch was trying to solve. 

Removing this branch as it is already out of date.

PR #435, opened 2022-02-03
Check that configured resource paths are not null

CJStadler:
Previously, the `String` from the configuration would always be passed to `Paths.get`. On openjdk >=16 passing `null` to `Paths.get` throws a `NullPointerException` (see https://bugs.openjdk.java.net/browse/JDK-8254876), so this would occur if a configuration option was missing. `additional.resources.location` was not included in the default `config.properties`, so the `NullPointerException` was always being thrown.

This change moves the `null` checks before `Paths.get` is called. It also adds `additional.resources.location` to the `config.properties`, mostly as an example since it is not strictly necessary.

I'm not able to run the test suite with openjdk 17 for unrelated reasons, but I did confirm that if I build from this branch and use that as a dependency in my application I can run it successfully with openjdk 17.

Resolves https://github.com/LinuxForHealth/hl7v2-fhir-converter/issues/427

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/resource/ResourceReader.java
@@ -80,17 +80,26 @@ public class ResourceReader {
    * @return The resource as a String
    */
   public String getResource(String resourcePath) {
-    Path resourceFolderFilePath = Paths.get(converterConfig.getResourceFolder(), resourcePath);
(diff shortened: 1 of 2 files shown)


cragun47:
@CJStadler Thank you for your contribution.  I'll merge it into master, and we'll pull it into the next release.  

We're glad you found this limitation.  We see the need to do this for OpenJDK 17,  however we're not ready to move to  full support for OpenJDK 17 because our devops environments don't yet support it (gradle and other dependencies.)

You will see some adjustments to our support statement that will be more limiting.

CJStadler:
Thanks so much for your help @cragun47! That makes sense re. OpenJDK 17.

PR #447, opened 2022-02-16
Fix Immun. only first OBX proccessed; javadoc

cragun47:

Fixes a bug where Immunization.fundingSource, Immunization.programEligibility worked by themselves, but not when they were together.  Applied the fix to similar Immunization elements.  Added test which checks for success when all are together.

Fixed some JavaDoc problems.


cragun47:
> Wondering if we should add some documentation about this somewhere as a tip?

Good suggestion.  I have added some documentation.

PR #468, opened 2022-03-02
Reuse converter through out tests

klwhaley:
Reuse converter instance through out tests.

Reduces junit run time.

Signed-off-by: KATHRYN L WHALEY <kathrynl@us.ibm.com>


cragun47 (approved):
No concerns.  Approved.

PR #469, opened 2022-03-10
Update build.gradle

snesm:
HAPI FHIR 5.1.0 update to 5.7.0


cragun47:
@snesm  Thank you for your contribution. Unfortunately I can't get it to pass the unit tests.  I'll have to look into why the failing test is failing on Windows, but not Unix.

snesm:
@cragun47 The build did pass on Windows once (not first time) before i completed the DCO (which started the process again).  I do not have a windows host available right now to test on, but this might be a possible cause: https://github.com/junit-team/junit5/issues/2811

cragun47:
@snesm  I agree that was the problem.  Running another time solved the build.   I will merge

PR #480, opened 2022-07-12
Updated ids to include nanoseconds prepended

klwhaley:


src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
```diff
@@ -134,6 +137,18 @@ public static String dateTimeWithZoneId(Object dateTime, Object zoneId) {
         return null;
     }
 
+    /** 
+     * Generates a string id for resource and bundle creation.
+     * Format is <current nano second>.<UUID>.
+     */
+    public static String generateResourceId() {
```
LisaWellman, line 144: Probably not critical to have the value from the same single point in time, but I think we should only get instant.now() once, and use the same result for both line 145 and line 147.

LisaWellman (approved):
Looks good, can you please create an issue to track this (I use them when creating the release notes)

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
klwhaley: fixed

PR #483, opened 2022-07-14
add process-timestamp

xpander53:


src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
```diff
@@ -137,6 +139,21 @@ public static String dateTimeWithZoneId(Object dateTime, Object zoneId) {
         return null;
     }
 
+    public static String getCurrentDateTimeUsingZoneId(Object zoneId) {
```
LisaWellman, line 142: Add javadoc for the method.
Include if the parm is required, or can be null (I think it can?).  

If it can't be null, should add a check for that

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
```diff
@@ -137,6 +139,21 @@ public static String dateTimeWithZoneId(Object dateTime, Object zoneId) {
...
+        // get the current data and time
+        LocalDateTime now = LocalDateTime.now();
+
+        // format now so it can be formatted by formatToDateTimeWithZone
+        final String FORMAT = "yyyyMMddHHmmss";
```
LisaWellman, line 147: move line 147-148 to be static variables at the class level (for performance reasons)

LisaWellman (approved):
Just a couple of small comments, approving

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
xpander53: Done

src/main/java/io/github/linuxforhealth/hl7/data/Hl7RelatedGeneralUtils.java
xpander53: Done

PR #484, opened 2022-07-20
initial checkin

xpander53:


src/test/java/io/github/linuxforhealth/hl7/segments/FHIRExtensionsTest.java
```diff
@@ -99,4 +106,61 @@ void testExtensionTwoRaces() {
 
     // See CodeableConceptText.java for more tests on CodeableConcepts
 
+    @Test
+    // Verifies the meta extension process-timestamp is present and verifies if the property
```
LisaWellman, line 110: Fix comment, it mentions `process-timestamp` but not `tenant-id`

LisaWellman (approved):
Looks good, just one code comment that needs updating

src/test/java/io/github/linuxforhealth/hl7/segments/FHIRExtensionsTest.java
xpander53: Done

PR #488, opened 2022-08-10
Issue 486 Remove dashes from SSNs

cragun47:
Signed-off-by: Brian Cragun <cragun@gmail.com>

This adds a DataValueResolved to remove dashes from SSNs.  It then uses that resolver where SSNs are found to remove the dashes.  Updates tests.


LisaWellman (approved):
Looks good

PR #492, opened 2022-09-12
Create codeql-analysis.yml

klwhaley:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
```
LisaWellman, line 1: Seems like we should remove this language, or perhaps better just add an intro comment on where this was copied from

LisaWellman (approved):

PR #504, opened 2023-08-18
add custom jexl functions

davidmurch:
I believe this is minimum change needed to add this capability. The implementer needs to create their own Engine extending  `HL7MessageEngine` here is an example

```Java
public class CustomHL7MessageEngine extends HL7MessageEngine {

    private Map<String, Object> customFunctions = new HashMap<>();
    
    public CustomHL7MessageEngine(FHIRContext context, Bundle.BundleType bundleType, Map<String, Object> customFunctions) {
        super(context, bundleType);
        this.customFunctions = customFunctions;
    }

    @Override
    public Bundle transform(InputDataExtractor dataInput, Iterable<FHIRResourceTemplate> resources, Map<String, EvaluationResult> contextValues) {
        HL7MessageData data = (HL7MessageData) dataInput;
        HL7MessageData newMessageData = new HL7MessageData(data.getHL7DataParser(), customFunctions);
        return super.transform(newMessageData, resources, contextValues);
    }
}
```

The role of the custom Engine is just to intercept the `transform` so the new constructor on HL7MessageData can be called.


stueynz:
@davidmurch and I are working on this project together - Test case will be there sometime today (NZ time)
We'll add a new CustomFunctions section to the TECHNIQUES.md - which seems the most appropriate place for a "HOW TO do wild stuff with custom functions"

LisaWellman:
@davidmurch  @stueynz I see you had another commit, please tag when you are ready for a review

LisaWellman:
@davidmurch @stueynz I see no activity here; do you still want to pursue?

davidmurch:
@LisaWellman 

Apologies for the delays as we completed our project using hl7v2-fhir-converter. I need to add our docs/examples etc and will update.

PR #506, opened 2023-09-07
Feature:  Ignore Empty HL7 Segments

stueynz:
Sometimes some segments are empty, and due to common templates populating `Resource.meta` field we need a way to ignore these empty segments, and not generate a partially filled in Resource.

This example message:

```
MSH|^~\\&|TestSystem||TestTransformationAgent||20150502090000||ADT^A09|controlID|P|2.6
EVN|A01|20150502090000|
PID|||1234^^^^MR||DOE^JANE^|||F||||||||||||||||||||||
PV1||I||||||||SUR||||||||S|VisitNumber^^^ACME|A||||||||||||||||||||||||20150502090000|
AL1|
```
with resource template setting of:

```yaml
resources:
    - resourceName: AllergyIntolerance
      segment: AL1
      resourcePath: resource/AllergyIntolerance
      repeats: true
      ignoreEmpty: true      ## We want to test our new ignoreEmpty flag
      additionalSegments:
        - MSH
```
Should not generate an `AllergyIntolerance` resource if the `ignoreEmpty` field is set to `true`

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/message/HL7FHIRResourceTemplateAttributes.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/message/HL7FHIRResourceTemplateAttributes.java
@@ -20,11 +20,11 @@ public class HL7FHIRResourceTemplateAttributes {
   private boolean repeats;
   private String resourcePath;
   private boolean isReferenced;
+  private boolean ignoreEmpty;
   private HL7Segment segment;// primary segment
   private List<HL7Segment> additionalSegments;
   private ResourceModel resource;
   private List<String> group;
-
 
 
   public HL7FHIRResourceTemplateAttributes(Builder builder) {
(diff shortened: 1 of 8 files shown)


src/main/java/io/github/linuxforhealth/hl7/message/HL7FHIRResourceTemplateAttributes.java
```diff
@@ -156,6 +160,11 @@ public Builder withIsReferenced(boolean isReferenced) {
       return this;
     }
 
+    public Builder withignoreEmpty(boolean ignoreEmpty) {
```
LisaWellman, line 163: This should be "withIgnoreEmpty" (capital 'I' on "Ignore" to have proper camel case.

stueynz: I've added a little commit - fixing this one.

LisaWellman:
@stueynz can you sign this one as well?  I see there is a build failure as well.

LisaWellman:
@stueynz I see no activity here; do you still want to pursue?

PR #515, opened 2023-11-28
Allowing for complex and or conditions with parens

JessicaWNava:
This PR allows for complex conditional statements
Test Steps:

Make sure the tests still pass

Changes
- Adds a compound conditional option

Checklist
Testing
[ X] Tested locally?


JessicaWNava:
@mkalish I think that maybe we could do A LOT of code cleanup here and use this logic for all of the cases rather than having conditionals with various types and stuff. Just don't want to go further down this rabbit hole if it is not Kalish approved. :)

Issue #512, opened 2023-10-20
StringIndexOutOfBoundsException is thrown while converting NTE segment with additional dash 


quake1997:
**Describe the bug**

When hl7 msg contains NTE segments with additional dash (-), `StringIndexOutOfBoundsException` exception is thrown while calling `HL7ToFHIRConverter::convert`. Issue related to changes implemented within https://github.com/LinuxForHealth/hl7v2-fhir-converter/pull/305/files#diff-c5d8954d3bc36408229772e082eb4fc36125a89c9e858fa22ec36aaa27c3c493R170

**To Reproduce**
```
@Test
void shouldCreateBundleResource() {
   HL7ToFHIRConverter ftv = new HL7ToFHIRConverter();
   String hl7message = "MSH|^~\\&||Test System|||20210917110100||OML^O21^OML_O21|||2.6\r"
                + "PID|1||7659afb9-0dfc-d744-1f40-5b9314807108^^^^MR||Feeney^Sam^^^^^L|||M||||||||\r"
                + "ORC|NW|8125550e-04db-11ec-a9a8-086d41d421ca^^ID^UUID||||||||||\r"
                + "OBR|1|8125550e-04db-11ec-a9a8-086d41d421ca^^ID^UUID||58410-2^CBC panel - Blood by Automated count^LN||||||||||||\r"
                + "NTE|1|L|Test Range:\r"
                + "NTE|2|L|Test1 and Test: 10 - 12\r"
                + "DG1|1||A013^Paratyphoid fever C^I10C|||A|||||||||1\r";
        
     assertNotNull(ftv.convert(hl7message));
 }
```

**Expected behavior**
 FHIR bundle resource is created correctly

PR #521, opened 2024-08-22
Issue/nte string index out of bound

ssharma5:
Fix for issue: StringIndexOutOfBoundsException is thrown while converting NTE segments with additional dash #512

In HL7ToFHIRConverter class, Method 'getHl7Message' has code for logging HL7 message structure. In order to optimize the content. It takes subset of each line based on first occurrence of "-" and 5 chars more. Issue occurs when there are multiple repeating segments like NTE and HL7Message.printStructure does not follow {Segment} - start prefix on these repeating segments. and these sgement has "-" in content.

For now added fail-safe check if indexof Dash + 5 chars is more then length of line then take line length as input.


ssharma5:
create new pull request with proper commit message.

PR #522, opened 2024-08-22
Fix for issue: StringIndexOutOfBoundsException is thrown while converting NTE segments with additional dash

ssharma5:
Fix for issue: StringIndexOutOfBoundsException is thrown while converting NTE segments with additional dash
[https://github.com/LinuxForHealth/hl7v2-fhir-converter/issues/512](url)

In HL7ToFHIRConverter class, Method 'getHl7Message' has code for logging HL7 message structure. In order to optimize the content. It takes subset of each line based on first occurrence of "-" and 5 chars more. Issue occurs when there are multiple repeating segments like NTE and HL7Message.printStructure does not follow {Segment} - start prefix on these repeating segments. and these sgement has "-" in content.

For now added fail-safe check if indexof Dash + 5 chars is more then length of line then take line length as input.

### DIFF
--- a/src/main/java/io/github/linuxforhealth/hl7/HL7ToFHIRConverter.java
+++ b/src/main/java/io/github/linuxforhealth/hl7/HL7ToFHIRConverter.java
@@ -185,7 +185,9 @@ public class HL7ToFHIRConverter {
                         output.append(line);
                     } else {
                         int firstDash = line.indexOf("-");
-                        output.append(line.substring(0, firstDash + 5));
+                        // Added fail-safe check if the content after "-" is less than 5 characters
+                        int lastContentIndex = Math.min(firstDash + 5, line.length());
+                        output.append(line.substring(0, lastContentIndex));
                     }
                     output.append("\n");
                 }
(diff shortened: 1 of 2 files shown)


tapeshmundhra (approved):
Approved

ssharma5:
@LisaWellman , need help on how can we proceed further to merge this fix.

LisaWellman:
Thanks for your contribution!

ssharma5:
Thanks @LisaWellman for supporting us on mering the PR. however i see one test is failing on windows CI, might be environment variable or settings issue. Also as we need to further propel this change into release. 
Is there any process for release. Also is there any chance we can get contributor access, so that we can debug and initiate this actions and releases as well. 
PS: As we are actively using this lib in our project we are expecting frequent changes mostly ading new config for new resource / fhir mappings and code fixes (if any).

LisaWellman:
@ssharma5 glad you are using it! I think that is reasonable - let me discuss with the other owners. It is vacation season so it may be a couple weeks, let me know if that is a problem with your project.

ssharma5:
Thanks @LisaWellman for consideration, It works for us, meanwhile we will work with local distribution.

Issue #524, opened 2025-04-16
Patient.telecom - ContactPoint.use and  ContactPoint.system  mappings from hl7v2 XTN datatype is incomplete


dnyanesh94:
ContactPoint.use shall be map from the XTN.2 (v2 table 0201)
ContactPoint.system  shall be map from the XTN.3 (v2 table 0202)

PR #525, opened 2025-04-16
#524 - Patient.telecom - ContactPoint.use and ContactPoint.system mappings from hl7v2 XTN datatype is incomplete

dnyanesh94:


ssharma5:
@dnyanesh94  please add DCO.

